PR #4, opened 2022-02-22
Update makefile tooling to run from vendor

JoelSpeed:
We aren't allowed to download anything while running code in OpenShift builds/CI (it forces `-mod=vendor` which prevents go install), so instead, we can just run the code from the vendor directory.


damdo (commented):
/lgtm

PR #5, opened 2022-02-28
Add ControlPlaneMachineSet beta API

JoelSpeed:
This is a part of the Machine v1beta1 API but will live here for now until we have completed the API review process. This will allow us to iterate on the project before we reach agreements on the API

This should allow us to start working on the logic of the operator without having to complete the API reviews first.
The API as is here matches the enhancement as of today.

Note, as this adds the first manifest for inclusion in the payload, and the operator depends on Machine API, it comes at level 31, 1 level after Machine API


pkg/api/machine/v1beta1/register.go
```diff
@@ -0,0 +1,40 @@
...
+	// SchemeGroupVersion generated code relies on this name
+	// Deprecated
+	SchemeGroupVersion = GroupVersion
+	// AddToScheme exists solely to keep the old generators creating valid code
+	// DEPRECATED
+	AddToScheme = schemeBuilder.AddToScheme
+)
+
+// Resource generated code relies on this being here, but it logically belongs to the group
+// DEPRECATED
```
alexander-demichev, line 25: why does is say "DEPRECATED" do want to replace/remove it?

alexander-demichev (approved):
/approve

pkg/api/machine/v1beta1/register.go
JoelSpeed: Copy and pasted this file from the machine v1beta1 package, good point, not needed

pkg/api/machine/v1beta1/types_controlplanemachineset.go
```diff
@@ -0,0 +1,245 @@
...
+	// ObjectMeta is the standard object metadata
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	// Labels are required to match the ControlPlaneMachineSet selector.
+	// +kubebuilder:validation:Required
+	ObjectMeta metav1.ObjectMeta `json:"metadata"`
+
+	// Spec contains the desired configuration of the Control Plane Machines.
+	// The ProviderSpec within contains platform specific details
+	// for creating the Control Plane Machines.
+	// The ProviderSe should be complete apart from the platform specific
```
lobziik, line 82: `ProviderSe` ?

JoelSpeed:
/override ci/prow/lint

Linting is coming in a later PR, I know this passes from local tests so this should be fine for now

PR #7, opened 2022-03-01
Add basic webhook set up

JoelSpeed:
This adds a validating webhook which will later make a number of validations against create and update events for ControlPlaneMachineSet resources.

At the moment, this is just very basic scaffolding for the webhooks, getting them registered and then boilerplate for the test suites.

Note, as the test requires a webhook manifest, I'm using the kubebuilder generator to build this and place it in the testdata folder close to the tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,12 @@ help: ## Display this help.
 
 ##@ Development
 
+.PHONY: manifests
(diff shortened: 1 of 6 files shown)


alexander-demichev (approved):
/approve

PR #13, opened 2022-03-14
Scaffold basic high level reconcile

JoelSpeed:
This PR adds a very high level reconcile skeleton of fetching the CPMS, setting the cluster operator status and updating the CPMS status. There is no real logic here but I've added some more test helpers and some tests to cover the ClusterOperator setting the status when there is no CPMS present, this logic should be implemented soon to ensure we can run the operator on a real cluster.

### DIFF
--- a/pkg/controllers/controlplanemachineset/cluster_operator_test.go
+++ b/pkg/controllers/controlplanemachineset/cluster_operator_test.go
@@ -1,0 +1,156 @@
+/*
+Copyright 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controlplanemachineset
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	configv1 "github.com/openshift/api/config/v1"
+	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test/resourcebuilder"
+	corev1 "k8s.io/api/core/v1"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var _ = Describe("Cluster Operator Status with a running controller", func() {
+	var mgrCancel context.CancelFunc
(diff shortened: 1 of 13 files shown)


pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	var namespaceName string
+
+	const operatorName = "control-plane-machine-set"
+
+	var co *configv1.ClusterOperator
+
+	BeforeEach(func() {
+		By("Setting up a namespace for the test")
+		ns := resourcebuilder.Namespace().WithGenerateName("control-plane-machine-set-cluster-operator-").Build()
```
damdo, line 48: Is the trailing `-` expected in the name for some reason or should it not be there?

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	var namespaceName string
+
+	const operatorName = "control-plane-machine-set"
+
+	var co *configv1.ClusterOperator
+
+	BeforeEach(func() {
+		By("Setting up a namespace for the test")
+		ns := resourcebuilder.Namespace().WithGenerateName("control-plane-machine-set-controller-").Build()
```
damdo, line 47: Same question from above for this.

damdo (commented):
Left a question, otherwise:
/lgtm

pkg/controllers/controlplanemachineset/cluster_operator_test.go
JoelSpeed: Yes, because the generate name will append a random 5 char string to the end, in which case, to ensure it's clear in any logging, I added a trailing `-` to separate the name prefix from the random suffix

damdo: Ok makes sense, thanks!

alexander-demichev (approved):
/approve

JoelSpeed:
/test unit

Once for luck

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -68,9 +72,57 @@ func (r *ControlPlaneMachineSetReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+	// Collect errors as an aggregate to return together after all patches have been performed.
+	var errs []error
+
+	result, err := r.reconcile(ctx, logger, cpms)
+	if err != nil {
+		// Don't return an error here so that we have an opportunity to update the status and cluster operator status.
+		errs = append(errs, fmt.Errorf("error reconciling control plane machine set: %w", err))
+	}
+
+	if err := r.updateControlPlaneMachineSetStatus(ctx, logger, cpms, patchBase); err != nil {
```
damdo, line 106: A question, given that this method is currently unimplemented. Are we planning to perform a client Patch (hence the patchBase) within it?
In that case would it be better to call it `patchControlPlaneMachineSetStatus`?

damdo (commented):
A question but otherwise
/lgtm

pkg/controllers/controlplanemachineset/controller.go
JoelSpeed: I had assume it would do a status patch yes, so, fair point, I don't have a strong opinion either way on the naming, whoever implements it can update the name as they please IMO

damdo: Ok sounds good 👍

JoelSpeed:
Manually added the lgtm as for some reason @damdo's comment didn't add the label

PR #15, opened 2022-03-16
Ensure the finalizer addition is the first action within the reconcile

JoelSpeed:
The finalizer is the only spec update we should be doing to the CPMS. So that we avoid conflicts, once we add the finalizer, we should make sure to requeue the object before attempting to reconcile the status in the next reconcile.

This sets up an `ensureFinalizer` function which will need implementing and the pending tests enabling.


alexander-demichev (approved):
/approve

elmiko (commented):
/lgtm

PR #16, opened 2022-03-23
Cleanup utils should not use default Gomega

JoelSpeed:
In order to make the cleanup util generic (and any future utils), we should make sure that the test that calls the util should be able to pass in their own Gomega. This will ensure that the util can be reused when a test declares its own Gomega (eg when using `NewWithT`). This is basically an effort to make sure this is generic enough to be copied into other projects/a shared library at some later date.


elmiko (commented):
nice upgrade
/approve

damdo (commented):
/lgtm

JoelSpeed:
/override ci/prow/e2e-aws

This isn't in payload yet, and even if it were, there's nothing actually being deployed, this code isn't running in the E2E yet

damdo (commented):
/lgtm

PR #17, opened 2022-03-24
Ensure deletion logic is handled when deletion timestamp is set

JoelSpeed:
When a deletion timestamp is set, we expect that the normal reconcile logic will be skipped. This forks the logic into a reconcileDelete function at the beginning. I've also added a couple of tests which outline what we are expecting from the deletion logic.


damdo (commented):
/lgtm

Fedosin (approved):
/approve

JoelSpeed:
/override ci/prow/e2e-aws

This isn't in payload yet, and even if it were, there's nothing actually being deployed, this code isn't running in the E2E yet

PR #18, opened 2022-03-24
Add conditions matchers for metav1.Condition

JoelSpeed:
As we are using metav1.Conditions in the ControlPlaneMachineSet itself, we also need conditions matchers for the metav1.Conditions. Posting this early to make the reviews easier on later PRs.

PR #19, opened 2022-03-25
Flesh out basic MachineProvider interface

JoelSpeed:
This PR outlines a MachineProvider interface that allows the core of the ControlPlaneMachineSet to gather information about the current state of the Machines and Nodes within the cluster, how these relate to different indexes (each Control Plane Machine replica is assigned an index in the set), and then make decisions on the rollouts based upon this information.

It is currently based on #18 so the first couple of commits should be ignored for review purposes.

This also now means we can start looking at how the status reporting will be implemented, I have added a number of test cases that will show how the status should look based on various states within the cluster.

The next PRs will flesh out a skeleton of the openshift machine API implementation of the provider and the skeleton of the update logic.


pkg/machineproviders/types.go
```diff
@@ -0,0 +1,83 @@
...
+	// This is used to inform the controller about decisions related to rolling out new machines.
+	NeedsUpdate bool
+
+	// Index denotes the Control Plane Machine index. Each Control Plane Machine replica is index (typically 0-2 in a
+	// three node cluster) and the Index will be needed to generate a replacement of this replica,  if a replacement is
+	// required.
+	Index int32
+
+	// ErrorMessage is used to provide information about any errors that have occurred with the Machine. For example, if
+	// the Machine has an error state within its status, it should be propogated up via this error message.
```
damdo, line 48:
```suggestion
	// the Machine has an error state within its status, it should be propagated up via this error message.
```

pkg/test/resourcebuilder/machine_info.go
```diff
@@ -0,0 +1,123 @@
...
+				Name: m.machineName,
+			},
+		}
+	}
+
+	if m.nodeName != "" {
+		info.NodeRef = &machineproviders.ObjectRef{
+			GroupVersionResource: m.nodeGVR,
+			ObjectMeta: metav1.ObjectMeta{
+				Name: m.machineName,
```
damdo, line 69:
```suggestion
				Name: m.nodeName,
```

pkg/machineproviders/types.go
```diff
@@ -0,0 +1,83 @@
...
+	// within the Cluster, as referred to by the ControlPlaneMachineSet.
+	// If there are any Machines missing (eg a failure domain is missing a Machine, or there are only 2 Machines when we
+	// expect 3), then a placeholder MachineInfo should be returned with enough information to create the Machine should
+	// that be required. It is up to the caller to decide how to handle the missing Machine.
+	GetMachineInfos() ([]MachineInfo, error)
+
+	// CreateMachine is used to instruct the Machine Provider to create a new Machine. The only input is the index for
+	// the new Machine. During construction of the MachineProvider, it should map indexes to failure domains so that it
+	// has all the required information for creating a new Machine stored, based solely on the index.
+	CreateMachine(int32) error
```
elmiko, line 77: i'm a little concerned about doing this all by index into an array, just curious if there is a way to be more definite?

or is there a reason why the indexing in preferable?

elmiko (commented):
in general looks really nice, i think the comment level is great. i just have a question about the design.

pkg/machineproviders/types.go
JoelSpeed: > i'm a little concerned about doing this all by index into an array

I tried to be very careful about the wording on the comment here, `it should map indexes to failure domains`, I'm not expecting an array but a map (which is kinda an array when you index here). The index is not an array index, but the index of the Control Plane Machine. Each Machine replica is assigned an index 0, 1, 2 today, we are going to keep using indexes going forward to identify the Machines, as you would see in a statefulset. This means when replica 0 needs replacement, the new machine comes up with a name that indicates it is also replica 0 (and should have the same failure domain).

The input to this function should allow the main reconciler to say, I want a new machine that replaces index x, and then, because the `MachineProvider` has already stored the config during initialisation, it knows index x goes into a certain failure domain and creates a new Machine with the correct config

alexander-demichev (approved):
/approve
I like the pluggable machine interface.

JoelSpeed:
/override ci/prow/e2e-aws

This isn't in payload yet and the AWS CI is experiencing an outage at the moment, these tests aren't currently needed

PR #21, opened 2022-03-29
Expand update strategies into their own reconcile functions

JoelSpeed:
This is a relatively small change to start forking the update strategy logic out. I've added some basic tests but will follow up with more tests for the RollingUpdate and OnDelete tests once this is merged. I expect those tests will be quite large so want to keep those separate for review purposes


pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -18,16 +18,68 @@ package controlplanemachineset
...
 	return ctrl.Result{}, nil
 }
+
+func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, machineProvider machineproviders.MachineProvider, indexedMachineInfos map[int32][]machineproviders.MachineInfo) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
+
+func (r *ControlPlaneMachineSetReconciler) reconcileMachineOnDeleteUpdate(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, machineProvider machineproviders.MachineProvider, indexedMachineInfos map[int32][]machineproviders.MachineInfo) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
```
elmiko, line 75: should there be a comment here to help direct what is required from these functions?

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -0,0 +1,201 @@
...
+				1: {i1m0, i1m1},
+				2: {i2m0},
+			},
+		}),
+		PEntry("all machines in the same index", tableInput{
+			in: []machineproviders.MachineInfo{i0m0, i0m1, i0m2},
+			expected: map[int32][]machineproviders.MachineInfo{
+				0: {i0m0, i0m1, i0m2},
+			},
+		}),
```
elmiko, line 199: just wondering if there is an error state we should be testing for?

elmiko (commented):
this mostly makes sense to me, just a couple questions

pkg/controllers/controlplanemachineset/updates_test.go
JoelSpeed: This function doesn't return any errors. It should be a simple filter function that takes the input and converts the format, from []MachineInfo to map[int32][]MachineInfo, so I don't think there's any error states

pkg/controllers/controlplanemachineset/updates.go
JoelSpeed: Good catch, missed those, will add those now

elmiko (commented):
updates look nice, thanks Joel
/approve

Fedosin (approved):
/lgtm

JoelSpeed:
/override ci/prow/e2e-aws

This repo isn't yet in payload and as such, these tests are not actually doing anything right now

PR #23, opened 2022-03-30
Fix "build" target in the Makefile

Fedosin:
Now "make build" fails with error:

```txt
go build -o bin/manager cmd/control-plane-machine-set-operator
cannot find package "." in:
        /home/mfedosin/projects/cluster-control-plane-machine-set-operator/vendor/cmd/control-plane-machine-set-operator
make: *** [Makefile:85: build] Error 1
```

This PR fixes it, allowing to successfully build the binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ verify-%: ## Ensure no diff after running some other target
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/control-plane-machine-set-operator
+	go build -o bin/manager ./cmd/control-plane-machine-set-operator
 
 define ensure-home
 	@ export HOME=$${HOME:=/tmp/kubebuilder-testing}; \


Makefile
```diff
@@ -82,7 +82,7 @@ verify-%: ## Ensure no diff after running some other target
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/control-plane-machine-set-operator
+	go build -o bin/manager cmd/control-plane-machine-set-operator/main.go
```
JoelSpeed, line 85: I've seen this before, and there is a different fix I think we concluded on.
There are implications of specifying the specific file IIRC.

Does `./cmd/control-plane-machine-set-operator` work?

JoelSpeed: I just tested my suggestion and it works, could you please change the PR to that? We saw issues before with the specific file approach so I'm hesitant to say that's what we should do here

Makefile
```diff
@@ -82,7 +82,7 @@ verify-%: ## Ensure no diff after running some other target
 
 .PHONY: build
 build: generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/control-plane-machine-set-operator
+	go build -o bin/manager .../cmd/control-plane-machine-set-operator
```
JoelSpeed, line 85: Should just be `./` shouldn't it? What's a triple dash?

damdo (commented):
/lgtm

PR #24, opened 2022-04-01
[OCPCLOUD-1496] Add tests for updating status of ControlPlaneMachineSet

JoelSpeed:
This PR adds some additional testing around the status updates to dictate how the log messages will be output when updates are and aren't required.

### DIFF
--- a/pkg/controllers/controlplanemachineset/status_test.go
+++ b/pkg/controllers/controlplanemachineset/status_test.go
@@ -29,9 +29,121 @@ import (
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test/resourcebuilder"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
 )
 
 var _ = Describe("Status", func() {
+	Context("updateControlPlaneMachineSetStatus", func() {
+		var namespaceName string
+		var logger test.TestLogger
+		var reconciler *ControlPlaneMachineSetReconciler
+		var cpms *machinev1.ControlPlaneMachineSet
+		var patchBase client.Patch
+
+		BeforeEach(func() {
+			By("Setting up a namespace for the test")
+			ns := resourcebuilder.Namespace().WithGenerateName("control-plane-machine-set-controller-").Build()
+			Expect(k8sClient.Create(ctx, ns)).To(Succeed())
+			namespaceName = ns.GetName()
+
+			By("Setting up the reconciler")
+			logger = test.NewTestLogger()
+			reconciler = &ControlPlaneMachineSetReconciler{
+				Namespace: namespaceName,
+				Scheme:    testScheme,
+			}
+
+			By("Setting up supporting resources")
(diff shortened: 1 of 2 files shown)


pkg/controllers/controlplanemachineset/status_test.go
```diff
@@ -29,9 +29,114 @@ import (
...
+			test.CleanupResources(Default, ctx, cfg, k8sClient, namespaceName,
+				&machinev1.ControlPlaneMachineSet{},
+			)
+		})
+
+		Context("when the status has changed", func() {
+			BeforeEach(func() {
+				cpms.Status.ObservedGeneration = 1
+				cpms.Status.Replicas = 2
+				cpms.Status.ReadyReplicas = 3
```
damdo, line 81: Aren't these not changing from what is set on the outer BeforeEach https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/24/files#diff-9929655c1e1817bd7ea9cd13da2ddf03af9f4543d9d910648898d49236eaef1cR62-R64

pkg/controllers/controlplanemachineset/status_test.go
```diff
@@ -29,9 +29,114 @@ import (
...
+					Message: updatingStatus,
+				}))
+			})
+		})
+
+		Context("when the status has not changed", func() {
+			BeforeEach(func() {
+				cpms.Status.ObservedGeneration = 2
+				cpms.Status.Replicas = 3
+				cpms.Status.ReadyReplicas = 4
```
damdo, line 113: Aren't these changing from what is set on the outer BeforeEach https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/24/files#diff-9929655c1e1817bd7ea9cd13da2ddf03af9f4543d9d910648898d49236eaef1cR62-R64

JoelSpeed: This one is changing, but that's deliberate, I added a comment above to explain why and the comment below on the patch base should also help add context

damdo: Ah gotcha! Thanks for adding the comment :)

damdo (commented):
/approve

pkg/controllers/controlplanemachineset/status_test.go
```diff
@@ -29,9 +29,116 @@ import (
...
+			}
+
+			By("Setting up supporting resources")
+			cpms = resourcebuilder.ControlPlaneMachineSet().WithNamespace(namespaceName).Build()
+			Expect(k8sClient.Create(ctx, cpms)).To(Succeed())
+
+			cpms.Status.ObservedGeneration = 1
+			cpms.Status.Replicas = 2
+			cpms.Status.ReadyReplicas = 3
+			Expect(k8sClient.Status().Update(ctx, cpms)).To(Succeed())
```
elmiko, line 65: i'm finding the use of replicas and readyreplicas to be non-obvious in these tests, is there a comment we could add here to help inform why we are manipulating the values?

elmiko (commented):
generally looks good, i just have one point of confusion

pkg/controllers/controlplanemachineset/status_test.go
JoelSpeed: Ahh right, these are just dummy values so that we can tell whether we are updating or not, I will add a comment

elmiko (commented):
/lgtm

PR #27, opened 2022-04-06
Update ControlPlaneMachineSet CRD 

Fedosin:
This PR:

1. Adds a script that syncs ControlPlaneMachineSet CRD with openshift/api repo
2. Bumps github.com/openshift/api
3. Syncs the CRD using the script


hack/crds-sync.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
JoelSpeed, line 1: Can we make sure to add a make target as in MAO so we can set up a CI check to check this is always synced?

JoelSpeed (commented):
Looks good, will need to check why vendor failed, one quick ask for the makefile so we can add a status check for PRs and then good to go

hack/crds-sync.sh
Fedosin: Sure! It's called `crds-sync` and included in the `test` target

Makefile
```diff
@@ -70,14 +70,18 @@ lint: ## Run golangci-lint over the codebase.
 	$(call ensure-home, ${GOLANGCI_LINT} run ./...)
 
 .PHONY: test
-test: generate fmt vet ## Run tests.
+test: generate fmt vet verify-crds-sync ## Run tests.
```
JoelSpeed, line 73: I'd drop this for now and have a separate verify job as as status check. The tests don't rely on everything being committed so this would break your ability to run the tests while hacking on it. If you can drop this line change I'm happy to approve and merge

Fedosin: done!

JoelSpeed (commented):
/lgtm
/approve

PR #33, opened 2022-04-21
[OCPCLOUD-1420] Implement status synchronization

Fedosin:
This commit allows to sync the operator status with ControlPlaneMachineSet by observing its conditions.


pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -77,9 +87,13 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
-		// CVO will create a blank cluster operator for us before the operator starts.
-		co = resourcebuilder.ClusterOperator().WithName(operatorName).Build()
-		Expect(k8sClient.Create(ctx, co)).To(Succeed())
+		_, err = reconciler.Reconcile(context.Background(), reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Name:      clusterControlPlaneMachineSetName,
+				Namespace: namespaceName,
+			},
+		})
+		Expect(err).To(Succeed())
```
JoelSpeed, line 96: The controller is already running within a manager, it should react to events we shouldn't need to call this manually

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -42,13 +46,18 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
 	BeforeEach(func() {
 		By("Setting up a namespace for the test")
 		ns := resourcebuilder.Namespace().WithGenerateName("control-plane-machine-set-cluster-operator-").Build()
 		Expect(k8sClient.Create(ctx, ns)).To(Succeed())
 		namespaceName = ns.GetName()
 
+		// CVO will create a blank cluster operator for us before the operator starts.
+		co = resourcebuilder.ClusterOperator().WithName(operatorName).Build()
+		Expect(k8sClient.Create(ctx, co)).To(Succeed())
+
```
JoelSpeed, line 60: I think if you move this back to where it was, you'll ensure the reconcile happens after the manager has started no? The controller should react to the object being created

pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -18,21 +18,135 @@ package controlplanemachineset
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // setClusterOperatorAvailable sets the control-plane-machine-set cluster operator status to available.
 // This is used primarily when a ControlPlaneMachineSet doesn't exist.
 func (r *ControlPlaneMachineSetReconciler) setClusterOperatorAvailable(ctx context.Context, logger logr.Logger) error {
-	return nil
+	co, err := r.getClusterOperator(ctx, logger)
+	if err != nil {
+		return err
```
JoelSpeed, line 36: You need to wrap this error

pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -18,21 +18,135 @@ package controlplanemachineset
...
+	return nil
+}
+
+// getClusterOperator returns an instance of Cluster Operator resource for control-plane-machine-set cluster operator.
+func (r *ControlPlaneMachineSetReconciler) getClusterOperator(ctx context.Context, logger logr.Logger) (*configv1.ClusterOperator, error) {
+	co := &configv1.ClusterOperator{}
+	err := r.Get(ctx, client.ObjectKey{Name: r.OperatorName}, co)
+
+	if err != nil {
+		err := fmt.Errorf("failed to get clusterOperator %q: %w", r.OperatorName, err)
```
JoelSpeed, line 120: This should be where we `return nil, err` and the logger should use keys and values

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -126,29 +145,42 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+						Name:      clusterControlPlaneMachineSetName,
+						Namespace: namespaceName,
+					},
+				})
+				Expect(err).To(Succeed())
 			})
 
-			PIt("Set the cluster operator available", func() {
-				co := resourcebuilder.ClusterOperator().WithName(operatorName).Build()
+			It("Set the cluster operator available", func() {
```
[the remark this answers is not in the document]
JoelSpeed: The conditions matcher is written in such a way that this is taken into account, notice there's a custom `MatchClusterOperatorStatusConditions`, you shouldn't need to worry about the timestamps

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -97,27 +111,32 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
-				{
-					Type:   configv1.OperatorDegraded,
-					Status: configv1.ConditionFalse,
-				},
-				{
-					Type:   configv1.OperatorUpgradeable,
-					Status: configv1.ConditionTrue,
-				},
-			})))
+		It("Set the cluster operator available", func() {
```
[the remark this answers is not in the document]
JoelSpeed: As below, this test didn't need to be updated, it covers the same thing and AFAICT, should pass now

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -128,29 +139,226 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+			Eventually(komega.UpdateStatus(cpms, func() {
+				conds := []metav1.Condition{
+					newCondition(conditionAvailable, metav1.ConditionTrue, reasonAllReplicasAvailable, ""),
+					newCondition(conditionProgressing, metav1.ConditionFalse, reasonAllReplicasUpdated, ""),
+					newCondition(conditionDegraded, metav1.ConditionFalse, reasonAsExpected, ""),
+				}
+				cpms.Status = resourcebuilder.ControlPlaneMachineSetStatusBuilder{}.WithConditions(conds).Build()
+			})).Should(Succeed())
+
+			Eventually(func() (bool, error) {
```
JoelSpeed, line 185: This should use the same pattern as the previous tests with the komega and MatchClusterOperatorStatusConditions

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -128,29 +139,226 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+				newCondition(conditionDegraded, metav1.ConditionFalse, reasonAsExpected, ""),
+			}
+
+			cpms.Status.Conditions = conds
+			Expect(reconciler.updateClusterOperatorStatus(context.Background(), logger.Logger(), cpms)).To(Succeed())
+
+			Expect(logger.Entries()).To(BeEmpty())
+		})
+
+		It("Update the status if conditions changed", func() {
```
JoelSpeed, line 323: Nit
```suggestion
		It("Updates the status if conditions have changed", func() {
```

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -128,29 +139,226 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+					Message: "Syncing cluster operator status",
+					KeysAndValues: []interface{}{
+						"available",
+						string(metav1.ConditionFalse),
+						"progressing",
+						string(metav1.ConditionFalse),
+						"degraded",
+						string(metav1.ConditionTrue),
+						"upgradable",
+						string(metav1.ConditionFalse),
```
JoelSpeed, line 346: Nit, this would be easier to read/understand as key value pairs on the same line
```suggestion
						"available", string(metav1.ConditionFalse),
						"progressing", string(metav1.ConditionFalse),
						"degraded", string(metav1.ConditionTrue),
						"upgradable", string(metav1.ConditionFalse),
```

JoelSpeed (commented):
I'd like to make sure the tests are consistent with the rest of the test suite, so I'd prefer to move back to the komega and HaveField/matchers style. Perhaps we can pair on the tests next week to resolve the comments

pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -18,21 +18,127 @@ package controlplanemachineset
...
+	}
+
+	if !needUpdate {
+		return nil
+	}
+
+	err := r.Status().Update(ctx, co)
+	if err != nil {
+		logger.Error(err, "failed to sync status for cluster operator")
+		return fmt.Errorf("failed to sync status for cluster operator %s, %w", r.OperatorName, err)
```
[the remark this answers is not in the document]
Fedosin: %s is better

pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -18,21 +18,127 @@ package controlplanemachineset
...
+	}
+}
+
+// syncStatus updates cluster operator status with given conditions.
+func (r *ControlPlaneMachineSetReconciler) syncStatus(ctx context.Context, logger logr.Logger, co *configv1.ClusterOperator, conds []configv1.ClusterOperatorStatusCondition) error {
+	// We need to perform update only if conditions have been changed.
+	needUpdate := false
+
+	for _, c := range conds {
+		if !v1helpers.IsStatusConditionPresentAndEqual(co.Status.Conditions, c.Type, c.Status) {
```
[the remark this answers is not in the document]
Fedosin: No, do you think it is necessary?

pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -18,21 +18,135 @@ package controlplanemachineset
...
+	return configv1.ClusterOperatorStatusCondition{
+		Type:               conditionType,
+		Status:             conditionStatus,
+		LastTransitionTime: metav1.Now(),
+		Reason:             reason,
+	}
+}
+
+// syncStatus updates cluster operator status with given conditions.
+func (r *ControlPlaneMachineSetReconciler) syncStatus(ctx context.Context, co *configv1.ClusterOperator, conds []configv1.ClusterOperatorStatusCondition) error {
```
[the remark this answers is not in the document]
Fedosin: now this function updates the status only if conditions have been changed.

pkg/controllers/controlplanemachineset/cluster_operator.go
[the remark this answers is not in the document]
JoelSpeed: Yes, I would expect that the message or reason may change through the code, and in this case, we should update with the latest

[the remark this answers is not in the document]
Fedosin: Done

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -128,29 +148,211 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+})
+
+var _ = Describe("Cluster Operator Status", func() {
+	const operatorName = "control-plane-machine-set"
+	var co *configv1.ClusterOperator
+	var reconciler *ControlPlaneMachineSetReconciler
+	var logger test.TestLogger
+	var namespaceName string
+
+	cpmsBuilder := resourcebuilder.ControlPlaneMachineSet().WithName(clusterControlPlaneMachineSetName).WithNamespace(namespaceName)
```
JoelSpeed, line 188: You need to construct this in the before each else the namespaceName is just the empty string
```suggestion
	var cpmsBuilder resourcebuilder.ControlPlaneMachineSetBuilder 
```
Then in the before each 
```
cpmsBuilder = resourcebuilder.ControlPlaneMachineSet().WithName(clusterControlPlaneMachineSetName).WithNamespace(namespaceName)
```

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -128,29 +148,211 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+		// CVO will create a blank cluster operator for us before the operator starts.
+		co = resourcebuilder.ClusterOperator().WithName(operatorName).Build()
+		Expect(k8sClient.Create(ctx, co)).To(Succeed())
 
+		logger = test.NewTestLogger()
+	})
+
+	AfterEach(func() {
+		test.CleanupResources(Default, ctx, cfg, k8sClient, namespaceName,
+			&configv1.ClusterOperator{},
```
JoelSpeed, line 211: You also need to delete the CPMSs
```suggestion
			&configv1.ClusterOperator{},
			&machinev1.ControlPlaneMachineSet{},
```

pkg/controllers/controlplanemachineset/cluster_operator_test.go
```diff
@@ -128,29 +148,211 @@ var _ = Describe("Cluster Operator Status with a running controller", func() {
...
+			expectedError      error
+			expectedLogs       []test.LogEntry
+		}
+
+		DescribeTable("should update the cluster operator status based on the ControlPlaneMachineSet conditions", func(in updateClusterOperatorStatusTableInput) {
+			originalCPMS := in.cpms.DeepCopy()
+
+			err := reconciler.updateClusterOperatorStatus(ctx, logger.Logger(), in.cpms)
+			if in.expectedError != nil {
+				Expect(err).To(MatchError(in.expectedError))
```
JoelSpeed, line 228: If there's an error, we don't check the conditions then right? I think we need a return after this
```suggestion
				Expect(err).To(MatchError(in.expectedError))
				return
```

pkg/test/resourcebuilder/control_plane_machine_set.go
```diff
@@ -120,3 +124,58 @@ func (m ControlPlaneMachineSetBuilder) WithStrategyType(strategy machinev1.Contr
...
 }
+
+// WithConditions sets the conditions for the controlplanemachineset builder.
+func (m ControlPlaneMachineSetBuilder) WithConditions(conditions []metav1.Condition) ControlPlaneMachineSetBuilder {
+	m.conditions = conditions
+	return m
+}
+
+// ControlPlaneMachineSetStatusCondition creates a new control plane machine set status condition builder.
+func ControlPlaneMachineSetStatusCondition() ControlPlaneMachineSetStatusConditionBuilder {
```
JoelSpeed, line 135: We can make this a generic conditions builder, as Machines also use metav1.Conditions. I'd recommend moving this to the conditions.go file and having a generic conditions build rather than calling it ControlPlaneMachineSetConditions

JoelSpeed (commented):
This is looking great now, a few small things but we are largely ready to merge now I think

PR #35, opened 2022-04-22
[OCPCLOUD-1496] Implement status reconciliation from MachineInfo

Fedosin:
This PR allows to update CPMS object status according to the data from MachineInfo.

### DIFF
--- a/pkg/controllers/controlplanemachineset/status.go
+++ b/pkg/controllers/controlplanemachineset/status.go
@@ -18,6 +18,7 @@ package controlplanemachineset
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	machinev1 "github.com/openshift/api/machine/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/controlplanemachineset/status.go
```diff
@@ -36,8 +37,25 @@ const (
 // updateControlPlaneMachineSetStatus ensures that the status of the ControlPlaneMachineSet is up to date after
 // the resource has been reconciled.
 func (r *ControlPlaneMachineSetReconciler) updateControlPlaneMachineSetStatus(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, patchBase client.Patch) error {
-	// TODO: Observe the data from the patchBase.Data() to determine if we actually need to update the status or not.
-	// Then log and update as per the test cases.
+	data, err := patchBase.Data(cpms)
+	if err != nil {
+		return fmt.Errorf("can't fetch patch data from CPMS object: %w", err)
```
JoelSpeed, line 42: I'd like to update the wording here
``` suggestion
		return fmt.Errorf("cannot calculate patch data from control plane machine set object: %w", err)
```

pkg/controllers/controlplanemachineset/status.go
```diff
@@ -36,8 +37,25 @@ const (
...
+	if string(data) == "{}" {
+		logger.V(3).Info(notUpdatingStatus)
+
+		return nil
+	}
+
+	err = r.Status().Update(ctx, cpms)
+	if err != nil {
+		return fmt.Errorf("failed to sync status for CPMS object: %w", err)
+	}
```
JoelSpeed, line 55: Inline the err declaration please, also change the wording
```suggestion
	if err := r.Status().Update(ctx, cpms); err != nil {
		return fmt.Errorf("failed to sync status for control plane machine set object: %w", err)
	}
```

JoelSpeed (commented):
I've left feedback on the base PR #33, please resolve that first, and then I've added some more feedback for the additional commit here

PR #36, opened 2022-04-29
Ensure log key regex handles single line declarations

JoelSpeed:
The previous regex could be greedy and include the closing bracket for a single line declaration with the previous capture group. This would mean it would always capture a second line and second closing bracket which broke the way the script determines there are invalid matches.


Fedosin (approved):
/approve
/lgtm

PR #37, opened 2022-05-09
Add context and logger to MachineProvider interface funcs

JoelSpeed:
This means that the context and logger can be passed in for each call to the various MachineProvider functions. Meaning the caller can customise the context and logger rather than having it pinned to the originals passed in when the MachineProvider is constructed.

This is a better pattern as we shouldn't be storing objects like contexts and loggers which have additional context added to them through different callers


elmiko (commented):
makes sense to me, although `gomock.Any()` seems like a spooky function lol
/lgtm

PR #38, opened 2022-05-10
Add skeleton of Machine Provider for OpenShift Machine v1beta1

JoelSpeed:
This PR fleshes out the basic MachineProvider skeleton for the OpenShift Machine v1beta1 implementation of the MachineProvider interface.

There are no tests for the implementation yet (Get/Create/Delete), these will follow in a separate PR along with a flesh out of more of the mapping and testing for the mapping.


pkg/machineproviders/providers/machineproviders_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+			BeforeEach(func() {
+				cpms := cpmsBuilder.Build()
+				cpms.Spec.Template.MachineType = machinev1.ControlPlaneMachineSetMachineType("invalid")
+
+				provider, err = NewMachineProvider(ctx, logger.Logger(), k8sClient, cpms)
+			})
+
+			It("returns an error", func() {
+				Expect(err).To(MatchError(fmt.Errorf("%w: invalid", errUnexpectedMachineType)))
```
damdo, line 68: I think this doesn't _exactly_ match this https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/38/files#diff-d6879f795ccd8a91161c5aff41dfae8f5a1cbb2a498852cab53c4f7f77f714fbR50

damdo (commented):
/lgtm
Just one small question

pkg/machineproviders/providers/machineproviders_test.go
JoelSpeed: Well the test passes 😅  Why do you think it wouldn't match?

damdo: Oh yes!
I thought the test was commented out like most of the others and I thought there was a mismatching `invalid` error message.
Turns out that `invalid` is the sample name of the non-existing provider you are specifying above.
Maybe worth putting it in a variable to make it clear it's referring to the same thing.

JoelSpeed:

This isn't yet in payload so cannot have caused any upgrade failures

PR #39, opened 2022-05-11
[OCPCLOUD-1543] Failure domain mapping

JoelSpeed:
This PR expands on how the mapping will work for the failure domains to Machine indexes.
This should give some structure and an outline of the behaviour through the tests that have been added.

Primarily it breaks the mapping down into three stages to be implemented:
- Enumerating the CPMS defined failure domains
- Enumerating the Machine defined failure domains
- Reconciling the two sets of failure domains making sure Machines take precedence


damdo (commented):
/lgtm

PR #40, opened 2022-05-13
[OCPCLOUD-1558] Extend provider config abstraction to be platform agnostic

JoelSpeed:
This adds the Inject, Extract, Equal and Raw functionality to the provider config abstraction to allow the core of the MachineProvider to be platform agnostic.

These 4 functions represent the needs of the MachineProvider on the provider config.

It will use InjectFailureDomain to take the base provider config and set the correct failure domain before converting it to raw and creating a new machine from it.

It will use ExtractFailureDomain to determine if the failure domain of a Machine is similar to a failure domain it has defined within the CPMS.

It will use Equal to determine if the desired provider config matches that of the Machines it gathers to work out whether they are in need of an update

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig_test.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig_test.go
@@ -25,6 +25,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test/resourcebuilder"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig_test.go
```diff
@@ -81,4 +82,181 @@ var _ = Describe("Provider Config", func() {
...
+						providerConfig: *resourcebuilder.AWSProviderSpec().WithAvailabilityZone("us-east-1a").Build(),
+					},
+				},
+				comparePC: &providerConfig{
+					platformType: configv1.AWSPlatformType,
+					aws: AWSProviderConfig{
+						providerConfig: *resourcebuilder.AWSProviderSpec().WithAvailabilityZone("us-east-1b").Build(),
+					},
+				},
+				expectedEqual: true,
```
damdo, line 228: Should this be equal?

damdo (commented):
/lgtm
with a question.

damdo (commented):
/lgtm

PR #41, opened 2022-05-17
Ensure Cluster ID label is set on Machine template

JoelSpeed:
This label is required by the Machine controller to be able to create machines, so is validated by the Machine webhook already. However, we also need the cluster ID to be able to name the Machines. Rather than making the user set this twice, re-use the existing API and ensure the value is set through webhook validation.


RadekManak:
/lgtm
Just the lint errors 😀

PR #43, opened 2022-05-18
Ensure master role and type label are present on OpenShift Machine Templates

JoelSpeed:
Various components within OpenShift expect the machine role label to be present to identify which Machines are Control Plane Machines.
We should enforce consistency here as, if we don't have those labels, other components may not operate correctly.


alexander-demichev (approved):
/lgtm

PR #44, opened 2022-05-23
[OCPCLOUD-1565] Ensure invalid cluster state prevents further reconciliation

JoelSpeed:
The first few commits in this PR are mainly refactor (and I recommend reviewing commit wise).

The main point of this PR is that there are a couple of scenarios (no ready machines, unmanaged control plane nodes) that should cause the operator to go degraded, we need to make sure this is accounted for.

This PR adds the skeleton for this and then adds tests to ensure the appropriate conditions are set given the scenarios

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -185,7 +185,8 @@ func (r *ControlPlaneMachineSetReconciler) reconcile(ctx context.Context, logger
 }
 
 // reconcileMachines uses the gathered machine info to set the status of the ControlPlaneMachineSet and then,
-// uses the machine provider to take appropriate actions to perform any requied roll outs.
+// after validating that the cluster state is as expected, uses the machine provider to take appropriate actions
+// to perform any requied roll outs.
 func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, machineProvider machineproviders.MachineProvider, machineInfos map[int32][]machineproviders.MachineInfo) (ctrl.Result, error) {
 	if err := reconcileStatusWithMachineInfo(logger, cpms, machineInfos); err != nil {
 		return ctrl.Result{}, fmt.Errorf("error reconciling machine info with status: %w", err)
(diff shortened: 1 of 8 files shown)


pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -249,3 +258,9 @@ func machineInfosByIndex(cpms *machinev1.ControlPlaneMachineSet, machineInfos []
 
 	return out, nil
 }
+
+// isControlPlaneMachineSetDegraded determines whether or not the ControlPlaneMachineSet
+// has a true, degraded condition.
+func isControlPlaneMachineSetDegraded(cpms *machinev1.ControlPlaneMachineSet) bool {
+	return false
+}
```
damdo, line 279: Should we mark this as TODO?

pkg/test/resourcebuilder/node.go
```diff
@@ -0,0 +1,91 @@
...
+package resourcebuilder
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	masterNodeRoleLabel = "node-role.kubernetes.io/master"
+	workerNodeRoleLabel = "node-role.kubernetes.io/master"
```
damdo, line 26:
```suggestion
	workerNodeRoleLabel = "node-role.kubernetes.io/worker"
```

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -236,6 +241,14 @@ func (r *ControlPlaneMachineSetReconciler) ensureOwnerReferences(ctx context.Con
 	return nil
 }
 
+// validateClusterState uses the machineInfos to validate that:
```
damdo, line 244: Should we mark this as a TODO?

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -167,6 +171,23 @@ func (r *ControlPlaneMachineSetReconciler) reconcile(ctx context.Context, logger
...
+	result, err := r.reconcileMachines(ctx, logger, cpms, machineProvider, indexedMachineInfos)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error reconciling machines: %w", err)
+	}
+
+	return result, nil
+}
+
+// reconcileMachines uses the gathered machine info to set the status of the ControlPlaneMachineSet and then,
+// after validating that the cluster state is as expected,  uses the machine provider to take appropriate actions
```
damdo, line 188:
```suggestion
// after validating that the cluster state is as expected, uses the machine provider to take appropriate actions
```

JoelSpeed (commented):
@damdo Thanks for the review, I've rebased this on main and fixed up the two suggestions you added. Decided against adding the TODOs as we have tests covering these parts of the logic which are pending, we won't lose track of the code to be done

damdo (commented):
Sounds good, thanks.
/lgtm

PR #45, opened 2022-05-25
[OCPCLOUD-1438] Implement validation webhook for CPMS resource

RadekManak:
This PR implements webhook for the CPMS resource. What's missing are tests for Azure, GCP and OpenStack platform because I did not want this PR to get any bigger.


pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+	}
+
+	// Ensure CPMS created with invalid name is not allowed
+	if cpms.Name != "cluster" {
+		return fmt.Errorf("invalid name %s: %w", cpms.Name, errControlPlaneInvalidName)
+	}
+
+	// Ensure required labels are set and all machines are matching the label selector
+	if err := checkMachineLabels(cpms); err != nil {
+		return err
```
JoelSpeed, line 113: Errors must be wrapped

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+	}
+
+	// Ensure failure domains of Control Plane Machines match the ControlPlaneMachineSet on create
+	if cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.FailureDomains.Platform != "" {
+		if err := checkFailureDomains(cpms, controlPlaneMachines); err != nil {
+			return err
```
JoelSpeed, line 119: Errors must be wrapped

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+	}
+
+	// Ensure Control Plane Machine count matches the ControlPlaneMachineSet replicas
+	if int(*cpms.Spec.Replicas) != len(controlPlaneMachines) {
+		return fmt.Errorf("replicas: %v, Control plane machines: %v: %w", *cpms.Spec.Replicas, len(controlPlaneMachines), errControlPlaneReplicasDoesNotMatchMachineCount)
+	}
+
+	// Ensure CPMS created with invalid name is not allowed
+	if cpms.Name != "cluster" {
+		return fmt.Errorf("invalid name %s: %w", cpms.Name, errControlPlaneInvalidName)
```
JoelSpeed, line 108: Likewise, reverse the order on this error message

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+			return err
+		}
+	}
+
 	return nil
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
 func (r *ControlPlaneMachineSetWebhook) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) error {
+	oldCPMS, ok := oldObj.(*machinev1.ControlPlaneMachineSet)
```
JoelSpeed, line 128: Soemtimes oldObj may be nil, have you covered against this?

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+
+	newCPMS, ok := newObj.(*machinev1.ControlPlaneMachineSet)
+	if !ok {
+		return errTypeMismatchCondition
+	}
+
+	// Ensure spec.Replicas is immutable on update
+	if *oldCPMS.Spec.Replicas != *newCPMS.Spec.Replicas {
+		return errControlPlaneReplicasChange
+	}
```
JoelSpeed, line 141: You need to check for nil

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+
+	// Ensure selector is immutable on update
+	if !reflect.DeepEqual(oldCPMS.Spec.Selector, newCPMS.Spec.Selector) {
+		return errControlPlaneLabelSelectorChange
+	}
+
+	// Ensure required labels are set and all machines are matching the label selector
+	if err := checkMachineLabels(newCPMS); err != nil {
+		return err
```
JoelSpeed, line 150: Error must be wrapped

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
 }
+
+// fetchControlPlaneMachines returns all control plane machines in the cluster.
+func (r *ControlPlaneMachineSetWebhook) fetchControlPlaneMachines(ctx context.Context) ([]*machinev1beta1.Machine, error) {
+	machineList := &machinev1beta1.MachineList{}
+	if err := r.c.List(ctx, machineList); err != nil {
+		return nil, fmt.Errorf("error querying api for machineList object: %w", err)
+	}
+
+	controlPlaneMachines := make([]*machinev1beta1.Machine, 0)
```
JoelSpeed, line 168: You don't need to use make here, this will suffice
```suggestion
	controlPlaneMachines := []*machinev1beta1.Machine{}
```

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+			controlPlaneMachines = append(controlPlaneMachines, machine)
+		}
+	}
+
+	return controlPlaneMachines, nil
+}
+
+// checkLabelSelectorMatchesTemplateLabel ensures that specified label is set on both machine template and selector with the same value.
+func checkLabelSelectorMatchesTemplateLabel(cpms *machinev1.ControlPlaneMachineSet, label string) error {
+	selector, ok := cpms.Spec.Selector.MatchLabels[label]
```
JoelSpeed, line 182: The selector may also use an expression, it's not guaranteed to use match labels. You should use the label package to check that the selector matches the labels map rather than checking explicitly for the presence of this particular label

For an example, we do this with MachineSets, https://github.com/openshift/machine-api-operator/blob/2e1274b68f997a9b40eaebde6f0a56d73d4542e6/pkg/webhooks/machineset_webhook.go#L177-L183

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+}
+
+// checkLabelSelectorMatchesTemplateLabels checks that the label selector matches the template labels.
+func checkLabelSelectorMatchesTemplateLabels(cpms *machinev1.ControlPlaneMachineSet) error {
+	for selectorLabel, selectorValue := range cpms.Spec.Selector.MatchLabels {
+		value, ok := cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.ObjectMeta.Labels[selectorLabel]
+		if !ok || selectorValue != value {
+			return errControlPlaneLabelSelectorDoesNotMatchMachine
+		}
+	}
```
JoelSpeed, line 206: As with previous comments, see how we do this with MachineSets, https://github.com/openshift/machine-api-operator/blob/2e1274b68f997a9b40eaebde6f0a56d73d4542e6/pkg/webhooks/machineset_webhook.go#L177-L183

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+
+	// Ensure Control Plane Machine role label is set
+	if err := checkLabelSelectorMatchesTemplateLabel(cpms, machineRoleLabelName); err != nil {
+		return fmt.Errorf("incorrect required label %s: %w", machineRoleLabelName, err)
+	}
+
+	// Ensure Control Plane Machine type label is set
+	if err := checkLabelSelectorMatchesTemplateLabel(cpms, machineTypeLabelName); err != nil {
+		return fmt.Errorf("incorrect required label %s: %w", machineTypeLabelName, err)
+	}
```
JoelSpeed, line 226: I think we can check that these are in the template, and then check the selector matches the labels, I don't think we need to check the selector for each of the individual labels

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+
+	// Ensure machines are matched by selectors
+	if err := checkLabelSelectorMatchesTemplateLabels(cpms); err != nil {
+		return err
```
JoelSpeed, line 230: Must wrap errors

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -89,24 +94,33 @@ var _ = Describe("Webhooks", func() {
...
 		It("with a valid spec", func() {
 			cpms := builder.Build()
 			Expect(k8sClient.Create(ctx, cpms)).To(Succeed())
 		})
 
-		PIt("with a disallowed name", func() {
+		It("with a disallowed name", func() {
 			cpms := builder.WithName("disallowed").Build()
-			Expect(k8sClient.Create(ctx, cpms)).To(MatchError(ContainSubstring("TODO")))
+			Expect(k8sClient.Create(ctx, cpms)).To(MatchError(ContainSubstring("control plane machine set name must be cluster")))
```
JoelSpeed, line 114: As you have defined an error constant, you can just `MatchError(<errVariable>)` and the matcher will unwrap the error until it finds the correct error

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -188,14 +202,44 @@ var _ = Describe("Webhooks", func() {
...
+			}
+			var usEast1aBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1a").WithSubnet(filterSubet)
+			var usEast1bBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1b").WithSubnet(filterSubet)
+			var usEast1cBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1c").WithSubnet(filterSubet)
+			var usEast1cBuilderWithSubnet = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1c").WithSubnet(filterSubetdifferent)
+			var usEast1cBuilderWithIDSubnet = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1c").WithSubnet(idSubnet)
+
+			var usEast1dBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1d").WithSubnet(filterSubet)
+			var usEast1eBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1e").WithSubnet(filterSubet)
+			var usEast1fBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1f").WithSubnet(filterSubet)
```
JoelSpeed, line 231: Nit, please be careful in using whitespace to create clean separations
```suggestion
			var filterSubet = machinev1.AWSResourceReference{
				Type: machinev1.AWSFiltersReferenceType,
				Filters: &[]machinev1.AWSResourceFilter{{
					Name:   "tag:Name",
					Values: []string{"aws-subnet-12345678"},
				},
			}}
				
			var filterSubetdifferent = machinev1.AWSResourceReference{
				Type: machinev1.AWSFiltersReferenceType,
				Filters: &[]machinev1.AWSResourceFilter{{
					Name:   "tag:Name",
					Values: []string{"aws-subnet-different"},
				},
			}}
			
			var idSubnet = machinev1.AWSResourceReference{
				Type: machinev1.AWSIDReferenceType,
				ID:   stringPtr("subnet-us-east-1c"),
			}
			
			var usEast1aBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1a").WithSubnet(filterSubet)
			var usEast1bBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1b").WithSubnet(filterSubet)
			var usEast1cBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1c").WithSubnet(filterSubet)
			var usEast1cBuilderWithSubnet = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1c").WithSubnet(filterSubetdifferent)
			var usEast1cBuilderWithIDSubnet = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1c").WithSubnet(idSubnet)
			var usEast1dBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1d").WithSubnet(filterSubet)
			var usEast1eBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1e").WithSubnet(filterSubet)
			var usEast1fBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1f").WithSubnet(filterSubet)
```

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -188,14 +202,44 @@ var _ = Describe("Webhooks", func() {
...
 
 			BeforeEach(func() {
+				test.CleanupResources(Default, ctx, cfg, k8sClient, namespaceName,
+					&machinev1beta1.Machine{},
+					&machinev1.ControlPlaneMachineSet{},
+				)
```
JoelSpeed, line 237: This should happen in an `AfterEach`, not in the `BeforeEach`

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -204,9 +248,16 @@ var _ = Describe("Webhooks", func() {
 
 				builder = resourcebuilder.ControlPlaneMachineSet().WithNamespace(namespaceName).WithMachineTemplateBuilder(machineTemplate)
 
+				var filterSubet = machinev1beta1.AWSResourceReference{
+					Filters: []machinev1beta1.Filter{{
+						Name:   "tag:Name",
+						Values: []string{"aws-subnet-12345678"},
+					},
+					}}
```
JoelSpeed, line 256: This file doesn't appear to be gofmt'd 🤔 
```suggestion
				var filterSubet = machinev1beta1.AWSResourceReference{
					Filters: []machinev1beta1.Filter{{
						Name:   "tag:Name",
						Values: []string{"aws-subnet-12345678"},
					},
				}}
```

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -17,11 +17,14 @@ limitations under the License.
 package failuredomain
```
JoelSpeed, line 17: Please make sure to update the tests for this package as there should be some tests for these parts of the code and the new parts that you're adding

JoelSpeed (commented):
This is a great start! Thanks for posting this. I've added a bunch of comments, mainly nits as it seems stylistically generally very good, one bigger thing though.

In other webhooks, we use the standard kubernetes field validation (https://pkg.go.dev/k8s.io/apimachinery/pkg/util/validation/field) package to produce errors that are the same format as with other webhooks and validations, I'd like us to use that here as well.

Examples of its usage can be found in the Machine API packages.

pkg/webhooks/controlplanemachineset/webhooks.go
RadekManak: Is correct behavior to reject when updating nil object?

pkg/webhooks/controlplanemachineset/webhooks_test.go
RadekManak: It did not work without `ContainSubstring`, because the errors are of different type, I think.  Replaced the literal with variable.

pkg/webhooks/controlplanemachineset/webhooks_test.go
RadekManak: This was to clean up objects from higher context. Restructured the tests to remove this.

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -33,7 +37,9 @@ const (
 var (
 	// errUnsupportedPlatformType is an error used when an unknown platform
 	// type is configured within the failure domain config.
-	errUnsupportedPlatformType = errors.New("unsupported platform type")
+	errUnsupportedPlatformType    = errors.New("unsupported platform type")
+	errMissingFailureDomain       = errors.New("missing failure domain configuration")
+	errMachineMissingProviderSpec = errors.New("machine is missing provider spec")
```
JoelSpeed, line 42: These need comments similar to the one above for unsupported platform type.

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -90,16 +155,197 @@ func NewFailureDomains(failureDomains machinev1.FailureDomains) ([]FailureDomain
...
+
+	return fd, nil
+}
+
+// newFailureDomainFromProviderSpecAWS creates a FailureDomain from AWS providerSpec.
+func newFailureDomainFromProviderSpecAWS(rawProviderSpec *runtime.RawExtension) (FailureDomain, error) {
+	providerSpec := machinev1beta1.AWSMachineProviderConfig{}
+	err := json.Unmarshal(rawProviderSpec.Raw, &providerSpec)
+
+	if err != nil {
```
JoelSpeed, line 266: Nit, inline the error
```suggestion
	providerSpec := machinev1beta1.AWSMachineProviderConfig{}
	
	if err := json.Unmarshal(rawProviderSpec.Raw, &providerSpec); err != nil {
```

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -90,16 +155,197 @@ func NewFailureDomains(failureDomains machinev1.FailureDomains) ([]FailureDomain
...
+
+	return NewAWSFailureDomain(awsFailureDomain), nil
+}
+
+// newFailureDomainFromProviderSpecAzure creates a FailureDomain from Azure providerSpec.
+func newFailureDomainFromProviderSpecAzure(rawProviderSpec *runtime.RawExtension) (FailureDomain, error) {
+	azureFailureDomain := machinev1.AzureFailureDomain{}
+	err := json.Unmarshal(rawProviderSpec.Raw, &azureFailureDomain)
+
+	if err != nil {
```
JoelSpeed, line 285:
```suggestion

	if err := json.Unmarshal(rawProviderSpec.Raw, &azureFailureDomain); err != nil {
```

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -90,16 +155,197 @@ func NewFailureDomains(failureDomains machinev1.FailureDomains) ([]FailureDomain
...
+
+	return NewGCPFailureDomain(gcpFailureDomain), nil
+}
+
+// newFailureDomainFromProviderSpecOpenStack creates a FailureDomain from OpenStack providerSpec.
+func newFailureDomainFromProviderSpecOpenStack(rawProviderSpec *runtime.RawExtension) (FailureDomain, error) {
+	openStackFailureDomain := machinev1.OpenStackFailureDomain{}
+	err := json.Unmarshal(rawProviderSpec.Raw, &openStackFailureDomain)
+
+	if err != nil {
```
JoelSpeed, line 309:
```suggestion
	
	if err := json.Unmarshal(rawProviderSpec.Raw, &openStackFailureDomain); err != nil {
```

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain_test.go
```diff
@@ -103,10 +104,96 @@ var _ = Describe("FailureDomains", func() {
...
+		Context("With AWS machines", func() {
+			var failureDomains []FailureDomain
+			var err error
+			providerSpec := resourcebuilder.AWSProviderSpec()
+			machines := []*machinev1beta1.Machine{}
+			for _, az := range []string{"us-east-1a", "us-east-1b", "us-east-1c"} {
+				ps := providerSpec.WithAvailabilityZone(az)
+				machines = append(machines, resourcebuilder.Machine().WithProviderSpecBuilder(ps).Build())
+
+			}
```
JoelSpeed, line 140: This should be in the before each so that it is fresh for each test, else they could (in theory) modify the setup and break following tests

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain_test.go
```diff
@@ -103,10 +104,96 @@ var _ = Describe("FailureDomains", func() {
...
+				))
+			})
+		})
+
+		Context("With invalid providerSpec", func() {
+			var failureDomains []FailureDomain
+			var err error
+			providerSpec := resourcebuilder.AWSProviderSpec()
+			machine := resourcebuilder.Machine().WithProviderSpecBuilder(providerSpec).Build()
+			machine.Spec.ProviderSpec.Value = nil
```
JoelSpeed, line 164: In the before each please

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain_test.go
```diff
@@ -103,10 +104,96 @@ var _ = Describe("FailureDomains", func() {
...
+			It("returns an nil list", func() {
+				Expect(failureDomains).To(BeNil())
+			})
+		})
+
+		Context("With an unsupported platform type", func() {
+			var failureDomains []FailureDomain
+			var err error
+			providerSpec := resourcebuilder.AWSProviderSpec()
+			machine := resourcebuilder.Machine().WithProviderSpecBuilder(providerSpec).Build()
```
JoelSpeed, line 183: In the before each please

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -40,12 +47,29 @@ const (
...
+	errControlPlaneReplicasDoesNotMatchMachineCount = errors.New("control plane machine set replicas field does not match number of control plane machines")
+	errControlPlaneReplicasChange                   = errors.New("control plane machine set replicas cannot be changed")
+	errReplicasNil                                  = errors.New("control plane machine set replicas field is nil")
+	errUpdateNilCPMS                                = errors.New("cannot update nil control plane machine set")
+	errControlPlaneLabelSelectorChange              = errors.New("control plane machine set label selector cannot be changed")
+	errFailureDomainsDoNotMatch                     = errors.New("control plane machine set failure domains do not match failure domains of machines")
+	errLabelNotSet                                  = errors.New("required label not set on machine template")
+	errUnknownMachineTemplateType                   = errors.New("unknown machine template type")
+	errSelectorDoesNotMatchTemplateLabels           = errors.New("control plane machine set selector does not match machine template labels")
+)
```
JoelSpeed, line 62: Each of these should have a comment expanding on when they are used

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -60,17 +68,212 @@ func (r *ControlPlaneMachineSetWebhook) SetupWebhookWithManager(mgr ctrl.Manager
...
+		return errTypeMismatchCondition
+	}
+
+	controlPlaneMachines, err := r.fetchControlPlaneMachines(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Ensure Control Plane Machine count matches the ControlPlaneMachineSet replicas
+	if int(*cpms.Spec.Replicas) != len(controlPlaneMachines) {
```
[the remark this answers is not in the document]
JoelSpeed: I think it's still necessary as there are ways that clients could set it and avoid the defaulting (though this is unsupported) which would cause issues here

pkg/webhooks/controlplanemachineset/webhooks.go
JoelSpeed: Hmm, not sure, I think it's possible to be nil and for it to still be a valid update call. Might have to play in cluster and see what happens.

Thinking it through, we have strict requirements for making sure some old and new things are the same, so we require the old object, if it's not set, then yes, we should error. Let's keep it and see if that breaks anything and revise later

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -62,15 +86,179 @@ var _ webhook.CustomValidator = &ControlPlaneMachineSetWebhook{}
...
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
 func (r *ControlPlaneMachineSetWebhook) ValidateDelete(ctx context.Context, obj runtime.Object) error {
 	return nil
 }
+
+// fetchControlPlaneMachines returns all control plane machines in the cluster.
+func (r *ControlPlaneMachineSetWebhook) fetchControlPlaneMachines(ctx context.Context) ([]*machinev1beta1.Machine, error) {
+	machineList := &machinev1beta1.MachineList{}
+	if err := r.client.List(ctx, machineList); err != nil {
+		return nil, fmt.Errorf("error querying api for machineList object: %w", err)
```
JoelSpeed, line 178: Nit, prefer this wording
```suggestion
		return nil, fmt.Errorf("error querying api for machines: %w", err)
```

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -84,118 +89,155 @@ var _ = Describe("Webhooks", func() {
...
-			machineTemplate = resourcebuilder.OpenShiftMachineV1Beta1Template().WithProviderSpecBuilder(providerSpec)
-			// Default CPMS builder should be valid, individual tests will override to make it invalid
-			builder = resourcebuilder.ControlPlaneMachineSet().WithNamespace(namespaceName).WithMachineTemplateBuilder(machineTemplate)
-		})
-
-		It("with a valid spec", func() {
-			cpms := builder.Build()
-			Expect(k8sClient.Create(ctx, cpms)).To(Succeed())
-		})
+		Context("when validating without without failure domains", func() {
```
JoelSpeed, line 92:
```suggestion
		Context("when validating without failure domains", func() {
```

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -84,118 +89,155 @@ var _ = Describe("Webhooks", func() {
...
-			}).WithMachineTemplateBuilder(
-				machineTemplate.WithLabels(map[string]string{
-					openshiftMachineRoleLabel:            masterMachineRole,
-					openshiftMachineTypeLabel:            masterMachineRole,
-					machinev1beta1.MachineClusterIDLabel: "different-id",
-				}),
-			).Build()
+			It("with a disallowed name", func() {
+				cpms := builder.WithName("disallowed").Build()
+				Expect(k8sClient.Create(ctx, cpms)).To(MatchError(ContainSubstring(errControlPlaneInvalidName.Error())))
```
JoelSpeed, line 115: This _should_ work, does it error?
```suggestion
				Expect(k8sClient.Create(ctx, cpms)).To(MatchError(errControlPlaneInvalidName))
```

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -84,118 +89,155 @@ var _ = Describe("Webhooks", func() {
...
-			var usEast1eBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1e")
-			var usEast1fBuilder = resourcebuilder.AWSFailureDomain().WithAvailabilityZone("us-east-1f")
+			var filterSubet = machinev1.AWSResourceReference{
+				Type: machinev1.AWSFiltersReferenceType,
+				Filters: &[]machinev1.AWSResourceFilter{{
+					Name:   "tag:Name",
+					Values: []string{"aws-subnet-12345678"},
+				}}}
+
+			var filterSubnetdifferent = machinev1.AWSResourceReference{
```
JoelSpeed, line 214: CamelCase
```suggestion
			var filterSubnetDifferent = machinev1.AWSResourceReference{
```

pkg/webhooks/controlplanemachineset/webhooks_test.go
JoelSpeed: This doesn't appear to be fixed

JoelSpeed (commented):
Did you get to review my previous comment https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/45#pullrequestreview-985854016?

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -47,13 +58,31 @@ type FailureDomain interface {
...
 type failureDomain struct {
 	platformType configv1.PlatformType
 
 	aws machinev1.AWSFailureDomain
+
+	azure machinev1.AzureFailureDomain
+
+	gcp machinev1.GCPFailureDomain
+
+	openStack machinev1.OpenStackFailureDomain
```
JoelSpeed, line 85: If you want you can squash these together a bit
```suggestion
	azure machinev1.AzureFailureDomain
	gcp machinev1.GCPFailureDomain
	openStack machinev1.OpenStackFailureDomain
```

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -76,12 +105,53 @@ func (f failureDomain) AWS() machinev1.AWSFailureDomain {
...
+
+// Equal compares the underlying failure domain.
+func (f failureDomain) Equal(other FailureDomain) bool {
+	if f.platformType != other.Type() {
+		return false
+	}
+
+	switch f.platformType {
+	case configv1.AWSPlatformType:
+		return reflect.DeepEqual(f.AWS(), other.AWS())
```
JoelSpeed, line 131: Why does AWS need deepequal but the others do not?

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain_test.go
```diff
@@ -103,10 +104,95 @@ var _ = Describe("FailureDomains", func() {
...
 			})
 
-			PIt("returns an empty list of failure domains", func() {
+			It("returns an empty list of failure domains", func() {
+				Expect(failureDomains).To(BeEmpty())
+			})
+		})
+	})
+
+	Context("NewFailureDomainsFromMachines", func() {
```
JoelSpeed, line 113: These tests are great, thanks! I would like to see tests for the Azure, GCP and OpenStack failure domains as well, but these can be followed up on a later PR, are you happy to own that as part of the failure domain cards for those platforms? It will likely involve writing some resourcebuilder code as well

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -62,15 +82,196 @@ var _ webhook.CustomValidator = &ControlPlaneMachineSetWebhook{}
...
+	if err != nil {
+		return fmt.Errorf("could not fetch existing control plane machines: %w", err)
+	}
+
+	// Ensure Control Plane Machine count matches the ControlPlaneMachineSet replicas
+	if cpms.Spec.Replicas == nil {
+		errs = append(errs, field.Required(field.NewPath("spec", "replicas"), "replicas field is required"))
+	} else if int(*cpms.Spec.Replicas) != len(controlPlaneMachines) {
+		errs = append(errs, field.Forbidden(field.NewPath("spec", "replicas"),
+			fmt.Sprintf("control plane machine set replicas (%d) does not match control plane machines (%d)", *cpms.Spec.Replicas, len(controlPlaneMachines))))
```
JoelSpeed, line 102: Nit: Just an update on the wording please
```suggestion
			fmt.Sprintf("control plane machine set replicas (%d) does not match the current number of control plane machines (%d)", *cpms.Spec.Replicas, len(controlPlaneMachines))))
```

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -62,15 +82,196 @@ var _ webhook.CustomValidator = &ControlPlaneMachineSetWebhook{}
...
 	return nil
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
 func (r *ControlPlaneMachineSetWebhook) ValidateDelete(ctx context.Context, obj runtime.Object) error {
 	return nil
 }
+
+// fetchControlPlaneMachines returns all control plane machines in the cluster.
+func (r *ControlPlaneMachineSetWebhook) fetchControlPlaneMachines(ctx context.Context) ([]*machinev1beta1.Machine, error) {
```
JoelSpeed, line 175: Just wondering if this list needs to be a list of machine pointers or if it can be just a list of machines, wdyt?

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -62,15 +82,196 @@ var _ webhook.CustomValidator = &ControlPlaneMachineSetWebhook{}
...
+	}
+
+	return errs
+}
+
+// checkFailureDomains ensures that failure domains of Control Plane Machines match the ControlPlaneMachineSet.
+func checkFailureDomains(cpms *machinev1.ControlPlaneMachineSet, controlPlaneMachines []*machinev1beta1.Machine) []error {
+	errs := []error{}
+	if cpms.Spec.Template.OpenShiftMachineV1Beta1Machine == nil {
+		return append(errs, field.Required(field.NewPath("spec", "template", "machines_v1beta1_machine_openshift_io"),
```
JoelSpeed, line 223: Just wondering if it would be better to create a constant for the path here and then use child, eg 

```
machineTemplatePath := field.NewPath("spec", "template", "machines_v1beta1_machine_openshift_io")
failureDomainsPath := machineTemplatePath.Child("failureDomains")
...
field.InternalError(failureDomainsPath.Child("platform")
```

If you were to do this across this file it may help to deduplicate sone of the path stuff, wdyt?

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -62,15 +82,196 @@ var _ webhook.CustomValidator = &ControlPlaneMachineSetWebhook{}
...
+	specifiedFailureDomains, err := failuredomain.NewFailureDomains(cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.FailureDomains)
+	if err != nil {
+		return append(errs, field.Invalid(field.NewPath("spec", "template", "machines_v1beta1_machine_openshift_io", "failureDomains"),
+			cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.FailureDomains,
+			fmt.Sprintf("error getting failure domains from control plane machine set machine template: %v", err)))
+	}
+
+	// All machines use one of specified failure domains
+	for _, missingFailureDomain := range missingFailureDomains(machineFailureDomains, specifiedFailureDomains) {
+		errs = append(errs, field.Forbidden(field.NewPath("spec", "template", "machines_v1beta1_machine_openshift_io", "failureDomains"), fmt.Sprintf("cluster machine is using unspecified failure domain %v", missingFailureDomain)))
```
JoelSpeed, line 246: How does this error come out, do we want to use the string representation maybe?

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -84,129 +90,176 @@ var _ = Describe("Webhooks", func() {
...
-			}).WithMachineTemplateBuilder(
-				machineTemplate.WithLabels(map[string]string{
-					openshiftMachineRoleLabel:            masterMachineRole,
-					openshiftMachineTypeLabel:            masterMachineRole,
-					machinev1beta1.MachineClusterIDLabel: "different-id",
-				}),
-			).Build()
+			It("with a disallowed name", func() {
+				cpms := builder.WithName("disallowed").Build()
+				Expect(apierrors.ReasonForError(k8sClient.Create(ctx, cpms))).To(BeEquivalentTo(`name: Invalid value: "disallowed": control plane machine set name must be cluster`))
```
JoelSpeed, line 116: Oh nice, good find!

Out of interest, why does it need a `BeEquivalentTo`?

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -233,17 +286,45 @@ var _ = Describe("Webhooks", func() {
...
 						usEast1aBuilder,
+						usEast1bBuilder,
+						usEast1cBuilderWithSubnet,
 					),
 				)).Build()
 
-				Expect(k8sClient.Create(ctx, cpms)).To(MatchError(ContainSubstring("TODO")))
+				err := k8sClient.Create(ctx, cpms)
+				Expect(err).To(MatchError(ContainSubstring(`Forbidden: cluster machine is using unspecified failure domain us-east-1c`)))
+				Expect(err).To(MatchError(ContainSubstring(`Forbidden: no cluster machine is using failure domain us-east-1c`)))
```
JoelSpeed, line 300: This could end up being confusing, as the messages look the same, wonder if we can improve that at all 🤔  Perhaps we need to change the string representation

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -265,7 +346,7 @@ var _ = Describe("Webhooks", func() {
 					),
 				)).Build()
 
-				Expect(k8sClient.Create(ctx, cpms)).To(MatchError(ContainSubstring("TODO")))
+				Expect(k8sClient.Create(ctx, cpms)).To(MatchError(ContainSubstring("Forbidden: no cluster machine is using failure domain us-east-1f")))
```
JoelSpeed, line 349: Can we make sure we are consistent with strings here, perhaps make all strings double quoted and just escape any internal quotes

JoelSpeed:
/approve
/lgtm

Looking at test coverage, the webhooks are looking good (83%) but the failure domain stuff is quite low (43%), let's make sure we pick that up at some point. Thanks for working on this

PR #50, opened 2022-06-29
Require Machine role and type to be master

JoelSpeed:
This follows up on #43 which introduced the requirement for the two labels. We had intended to make sure that the label and values were as we expect, but we just made sure the label itself was present.

While reviewing https://issues.redhat.com/browse/OCPCLOUD-1438 and the test cases introduced, I realised we had missed the check for the value being something other than `master`, this PR should resolve that issue.


Fedosin (approved):
/lgtm

JoelSpeed:
/override ci/prow/e2e-aws

Failures are unrelated as this isn't in payload yet

PR #51, opened 2022-06-29
Refactor table tests to use cpmsBuilder, stick to camelCase log key convention

damdo:
- **refactor table tests to use cpmsBuilder**
   using the cpms pointer was causing issues with not retaining the passed
in values.
- **use lowercase keys for logging**

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller_test.go
+++ b/pkg/controllers/controlplanemachineset/controller_test.go
@@ -654,14 +654,15 @@ var _ = Describe("machineInfosByIndex", func() {
 	i2m0 := resourcebuilder.MachineInfo().WithIndex(2).WithMachineName("machine-2-0").Build()
 
 	type tableInput struct {
-		cpms          *machinev1.ControlPlaneMachineSet
+		cpmsBuilder   resourcebuilder.ControlPlaneMachineSetBuilder
 		machineInfos  []machineproviders.MachineInfo
 		expected      map[int32][]machineproviders.MachineInfo
 		expectedError error
 	}
 
 	DescribeTable("should sort Machine Infos by index", func(in tableInput) {
-		out, err := machineInfosByIndex(in.cpms, in.machineInfos)
+		cpms := in.cpmsBuilder.Build()
+		out, err := machineInfosByIndex(cpms, in.machineInfos)
 		if in.expectedError != nil {
 			Expect(err).To(MatchError(in.expectedError))
 			return
@@ -676,11 +677,11 @@ var _ = Describe("machineInfosByIndex", func() {
 		}
 	},
 		Entry("with no replicas in the ControlPlaneMachineSet", tableInput{
-			cpms:          &machinev1.ControlPlaneMachineSet{},
+			cpmsBuilder:   resourcebuilder.ControlPlaneMachineSet().WithReplicas(0),
 			expectedError: errReplicasRequired,
 		}),
 		Entry("no machine infos with 3 replicas", tableInput{
-			cpms:         resourcebuilder.ControlPlaneMachineSet().WithReplicas(3).Build(),
+			cpmsBuilder:  resourcebuilder.ControlPlaneMachineSet().WithReplicas(3),
 			machineInfos: []machineproviders.MachineInfo{},
 			expected: map[int32][]machineproviders.MachineInfo{
 				0: {},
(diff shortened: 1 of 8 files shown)


pkg/test/resourcebuilder/control_plane_machine_set.go
```diff
@@ -23,17 +23,29 @@ import (
...
 
 const (
-	// controlPlaneMachineSetName is the only valid name allowed.
+	// ControlPlaneMachineSetName is the only valid name allowed.
 	// A ControlPlaneMachineSet is a singleton within the cluster, this matches other singletons such as Infrastructure.
-	controlPlaneMachineSetName = "cluster"
+	ControlPlaneMachineSetName = "cluster"
 )
 
+type CPMS interface {
```
JoelSpeed, line 31: Acronyms are bad, lets spell this out and call it (common pattern from kube). Make sure to add godocs in this file too
```suggestion
type ControlPlaneMachineSetInterface interface {
```

pkg/test/resourcebuilder/control_plane_machine_set.go
```diff
@@ -23,17 +23,29 @@ import (
...
+	Build() *machinev1.ControlPlaneMachineSet
+}
+
+type NakedControlPlaneMachineSet struct {
+	Cpms *machinev1.ControlPlaneMachineSet
+}
```
JoelSpeed, line 37: Lets make this `Raw` like this, again, common naming we see in other places in kube
```suggestion
type RawControlPlaneMachineSet struct {
	Raw *machinev1.ControlPlaneMachineSet
}
```

pkg/test/resourcebuilder/control_plane_machine_set.go
```diff
@@ -23,17 +23,29 @@ import (
...
+
+func (m NakedControlPlaneMachineSet) Build() *machinev1.ControlPlaneMachineSet {
+	return m.Cpms
```
JoelSpeed, line 40: Return a copy of the CPMS so that this can be reused like the other builder which builds a unique CPMS each time

damdo: Yep, good point!

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -676,11 +677,30 @@ var _ = Describe("machineInfosByIndex", func() {
...
 		Entry("with no replicas in the ControlPlaneMachineSet", tableInput{
-			cpms:          &machinev1.ControlPlaneMachineSet{},
+			cpmsBuilder: resourcebuilder.RawControlPlaneMachineSet{Raw: &machinev1.ControlPlaneMachineSet{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       resourcebuilder.ControlPlaneMachineSetName,
+					Namespace:  resourcebuilder.OpenshiftMachineAPINamespaceName,
+					Generation: 1,
+				},
+				Spec: machinev1.ControlPlaneMachineSetSpec{
+					Replicas: nil,
```
JoelSpeed, line 687: It might be worth adding a comment to explain why we are using the raw builder in this case?

pkg/test/resourcebuilder/control_plane_machine_set.go
```diff
@@ -23,17 +23,30 @@ import (
...
+}
+
+type RawControlPlaneMachineSet struct {
+	Raw *machinev1.ControlPlaneMachineSet
+}
+
+func (m RawControlPlaneMachineSet) Build() *machinev1.ControlPlaneMachineSet {
+	cpms := m.Raw.DeepCopy()
+	return cpms
+}
```
JoelSpeed, line 42: Godoc missing on these

JoelSpeed (commented):
I like the interface and I think it's useful to be able to have a Raw option.

The other alternative is to have a `funcs` implementation, which would look something like

```
type CPMSFuncs struct {
  BuildFunc func() *machinev1.ControlPlaneMachineSet
}

func (c *CPMSFuncs) Build() *machinev1.ControlPlaneMachineSet {
  return c.BuildFunc()
}

...

// Then being used in tests it would be

cpmsBuilder: resourcebuilder.CPMSFuncs{BuildFunc: func() *machinev1.ControlPlaneMachineSet{
  cpms := cpmsBuilder.Build() // re-use existing set up
  cpms.Spec.Replicas = nil // This is what we are testing in this case
  return cpms
}},
```
I'm wondering if this funcs approach allows a more succinct and obvious test variation in the specific case we are trying to handle here?

PR #52, opened 2022-07-01
[OCPCLOUD-1587] Validate discriminated unions via webhooks

JoelSpeed:
This PR is made up of 2 commits. The first of which is a prefactor to make the second commit easier to work with.

The new content here is a set of validations added so that discriminated unions are validated correctly. IE, when the discriminant is set, only the correct field within the union is set, no extras. This should prevent users from setting invalid configurations

### DIFF
--- a/pkg/webhooks/controlplanemachineset/webhooks.go
+++ b/pkg/webhooks/controlplanemachineset/webhooks.go
@@ -48,6 +48,10 @@ const (
 	// masterMachineRole is the master role/type that is required to be set on
 	// all OpenShift Machine API Machine templates.
 	masterMachineRole = "master"
+
+	// clusterSingletonName is the OpenShift standard name, "cluster", for singleton
+	// resources. All ControlPlaneMachineSet resources must use this name.
+	clusterSingletonName = "cluster"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


damdo (commented):
Nice refactor, just a non-blocking question on naming.
Additions around discriminated union validation look good.
/lgtm

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -363,3 +404,66 @@ func missingFailureDomains(list1 []failuredomain.FailureDomain, list2 []failured
...
+			errs = append(errs, field.Forbidden(parentPath.Child(fieldJSONName), fmt.Sprintf("value not allowed when %s is %q", discriminantJSONName, discriminantValue)))
+		}
+	}
+
+	return errs
+}
+
+// getJSONName gets the JSON name of the field from the struct tag.
+// This is used so that errors show the name as the end user would use.
+func getJSONName(typ reflect.StructField) string {
```
alexander-demichev, line 467: what does `typ` mean?

JoelSpeed: Yeah good point, I'm gonna rename this to `structField`

damdo (commented):
/lgtm

alexander-demichev (approved):
/lgtm

JoelSpeed:

This PR doesn't affect the E2Es yet as this isn't in payload

PR #53, opened 2022-07-04
tests: use builders for logs and errors tableinputs

damdo:
It updates the tests TableInput to use builders for errors and logs, to allow computation of parametrized input values within the test scope.


damdo:
I was erring on the consistency side here, that's why I changed it everywhere yes.

I agree it slightly more verbose but I'm not sure is that much more complex in itself.

It might take a non negligible amount of time to find out which suites will really need the builder funcs, as every test entry would need to be searched for parametrised values.
It would be much easier instead to do it further down the line when the tests are "Un-Pended" and they are actually testing things.

Not sure about the alternative. I think it can be a bit confusing for the user to specify both and know which one takes precedence.

For now I can change only the *update* test suite. Then when the other suites will test against an actual implementation, we'll be able to find out which needs these changes and change them accordingly. WDYT?

JoelSpeed:
> For now I can change only the update test suite. Then when the other suites will test against an actual implementation, we'll be able to find out which needs these changes and change them accordingly. WDYT?

Yep go ahead, change the ones that are causing you issues and we can resolve the rest later, probably worth keeping the commit as is now around in case we do need it later

damdo:
>> For now I can change only the update test suite. Then when the other suites will test against an actual implementation, we'll be able to find out which needs these changes and change them accordingly. WDYT?

> Yep go ahead, change the ones that are causing you issues and we can resolve the rest later, probably worth keeping the commit as is now around in case we do need it later

Done

damdo:
This PR changes only unit tests and doesn't affect e2e behaviour.
Skipping e2e.

/override ci/prow/e2e-aws

damdo:
This PR changes only unit tests and doesn't affect e2e behaviour.
Skipping e2e.

/override ci/prow/e2e-aws

PR #54, opened 2022-07-04
[OCPCLOUD-1586] Update ControlPlaneMachineSet API from openshift/api

JoelSpeed:
This updates the API based on https://github.com/openshift/api/pull/1216.
There were a few changes to the API made within that PR that need to be accounted for here, mainly the changes of the AWS resource reference discriminator values and then some clarification of the status fields


damdo (commented):
/lgtm

PR #55, opened 2022-07-04
Implement finalizer addition

RadekManak:
This PR implements finalizer addition for CPMS and fixes nil dereference with nil provider spec.

I don't quite understand why handling stale CPMS is required, but hope it's implemented correctly.


pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -123,7 +129,7 @@ func (r *ControlPlaneMachineSetReconciler) Reconcile(ctx context.Context, req ct
 	}
 
 	// Take a copy of the original object to be able to create a patch for the status at the end.
-	patchBase := client.MergeFrom(cpms)
+	patchBase := client.MergeFrom(cpms.DeepCopy())
```
RadekManak, line 132: This one took me way too long to find.

PR #56, opened 2022-07-04
[OCPCLOUD-1502] RollingUpdate Strategy

damdo:
This introduces an implementation of the `RollingUpdate` Strategy for the ControlPlaneMachineSet Machines.

### DIFF
--- a/pkg/controllers/controlplanemachineset/updates.go
+++ b/pkg/controllers/controlplanemachineset/updates.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sort"
 
 	"github.com/go-logr/logr"
 	machinev1 "github.com/openshift/api/machine/v1"
@@ -54,6 +55,11 @@ const (
 	// noUpdatesRequired is a log message used to inform the user that no updates are required within
 	// the current set of Machines.
 	noUpdatesRequired = "No updates required"
+
+	// noCapacityForExpansion is a log message used to inform the user that no capacity for expanding the machine count
+	// by creating a new machine is left as maximum surge has been reached.
+	noCapacityForExpansion = "Insufficient capacity for expansion, maximum surge has been reached." +
+		" Cannot create a replacement Machine at this time."
 
 	// removingOldMachine is a log message used to inform the user that an old Machine has been
 	// deleted as a part of the rollout operation.
@@ -137,6 +143,49 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
 // In certain scenarios, there may be indexes with missing Machines. In these circumstances, the update should attempt
 // to create a new Machine to fulfil the requirement of that index.
 func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, machineProvider machineproviders.MachineProvider, indexedMachineInfos map[int32][]machineproviders.MachineInfo) (ctrl.Result, error) {
+	logger = logger.WithValues("updateStrategy", cpms.Spec.Strategy.Type)
+
+	// To ensure an ordered and safe reconciliation,
+	// one index at a time is considered.
+	// Indexes are sorted in ascendent order, so that all the operations of the same importance,
+	// are executed prioritizing the lower indexes first.
+	sortedIndexedMs := sortMachineInfosByIndex(indexedMachineInfos)
+
+	// The maximum number of machines that
+	// can be scheduled above the original number of desired machines.
+	// At present, the surge is limited to a single Machine instance.
+	// NOTE: If this gets changed or parametrized,
+	// the tests will need to be updated accordingly.
+	maxSurge := 1
+	// Devise the existing surge and keep track of the current surge count.
+	// No check for early stoppage is done here,
+	// as deletions can continue even if the maxSurge has been already reached.
+	surgeCount := deviseExistingSurge(cpms, sortedIndexedMs)
+
+	var updated bool
+
+	for idx, machines := range sortedIndexedMs {
+		if done, result, err := r.deleteReplacedMachines(ctx, logger, machineProvider, machines); err != nil {
+			return result, err
+		} else if done {
+			updated = true
+		}
+
+		if done := r.waitForPendingMachines(logger, machines); done {
+			updated = true
+		}
+
+		if done, result, err := r.createReplacementMachines(ctx, logger, machineProvider, machines, idx, maxSurge, &surgeCount); err != nil {
+			return result, err
+		} else if done {
+			updated = true
+		}
+	}
+
+	if !updated {
+		logger.V(4).Info(noUpdatesRequired)
+	}
+
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +139,85 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+			// Wait for it to become Ready.
+			logger = logger.WithValues("index", idx, "namespace", r.Namespace, "name", getMachineReplacementName(idx))
+			logger.V(2).Info(waitingForReady)
+			return ctrl.Result{}, nil
+		}
+	}
+
+	// Reconcile machines that need an update.
+	for idx, machines := range sortedIndexedMs {
+		// Find out if and what Machines in this index need an update.
```
JoelSpeed, line 180: Might be worth breaking the content of this loop into a helper function

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +139,85 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+	// Indexes are sorted in ascendent order, so that all the operations of the same importance,
+	// are executed prioritizing the lower indexes first.
+	sortedIndexedMs := sortMachineInfos(indexedMachineInfos)
+
+	// The maximum number of machines that
+	// can be scheduled above the original number of desired machines.
+	// At present, the surge is limited to a single Machine instance.
+	maxSurge := 1
+	// Devise the existing surge and keep track of the current surge count.
+	surgeCount := deviseExistingSurge(cpms, sortedIndexedMs)
```
JoelSpeed, line 155: What happens if the surgeCount is already met? Should there be an early exit here or is this handled later? 

I think the answer is no because we need to handle deletes, but perhaps there should be a comment to stop people like me asking why we don't do it here

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +233,134 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+
+// getMachineName returns the canonical name for a machine of a certain index.
+func getMachineName(idx int) string {
+	return "machine-" + strconv.Itoa(idx)
+}
+
+// getMachineReplacementName returns the canonical name for a replacement machine of a certain index.
+func getMachineReplacementName(idx int) string {
+	return "machine-replacement-" + strconv.Itoa(idx)
+}
```
JoelSpeed, line 296: Didn't expect to see these here, where are they being used? I would expect to be getting names from the MachineInfo not from a function like this, the names will vary per machine right?

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +233,134 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+
+// sortMachineInfos returns a sorted list of indexes together with their MachineInfos.
```
JoelSpeed, line 298: Sort by what?

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +233,134 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
 	return ctrl.Result{}, nil
 }
+
+// isDeletedMachine checks if a machine is deleted.
+func isDeletedMachine(m machineproviders.MachineInfo) bool {
+	return m.MachineRef.ObjectMeta.DeletionTimestamp != nil
+}
+
+// needUpdateMachines returns the list of MachineInfo which have Machines that need an update.
+func needUpdateMachines(machinesInfo []machineproviders.MachineInfo) []machineproviders.MachineInfo {
```
JoelSpeed, line 243: There's a lot of repeated looping here, what if you had a `filterMachines` function that took filters as an argument that could then be composed, would this make it clearer or more complicated?

```
type filterMachines func(machineprovider.MachineInfo) bool

func filterMachines(machinesInfo []machineproviders.MachineInfo, filters ...filterFunc) []machineproviders.MachineInfo {
  filtered := []machineproviders.MachineInfo{}

  for _, m := range machineInfos {
    for _, filter := range filters {
      if !filter(m) {
        continue
      }
    }

    filtered = append(filtered, m)
  }

  return filtered
}
```

Then each of your other helpers can become a combination of filters into this 🤔 

```
func not(filter filterFunc) filterFunc {
  return func(m machineproviders.MachineInfo) bool {
    return !filter(m)
  }
}

func machineNeedsUpdate(m machineproviders.MachineInfo) bool {
  return m.NeedsUpdate
}

func machineIsReady(m machineproviders.MachineInfo) bool {
  return m.Ready
}

func pendingMachines(machinesInfo []machineproviders.MachineInfo) []machineproviders.MachineInfo {
  return filterMachines(machinesInfo, not(machineIsReady), not(machineNeedsUpdate))
}

func updatedMachines(machinesInfo []machineproviders.MachineInfo) []machineproviders.MachineInfo {
  return filterMachines(machinesInfo, machineIsReady, not(machineNeedsUpdate))
}

func readyMachines(machinesInfo []machineproviders.MachineInfo) []machineproviders.MachineInfo {
  return filterMachines(machinesInfo, machineIsReady)
}

```

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +139,85 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+			if any(pendingMachines(machines)) {
+				// A Pending Machine Replacement is present.
+				// Wait for it to become Ready.
+				logger.V(2).WithValues("replacementName", getMachineReplacementName(idx)).Info(waitingForReplacement)
+				return ctrl.Result{}, nil
+			}
+
+			// No Healthy or Pending Replacement Machine exists,
+			// trigger a Machine creation.
+			return createMachine(ctx, logger, machineProvider, idx, maxSurge, &surgeCount)
```
JoelSpeed, line 214: Handling the surge inside another function means that, in the future, if we need to create multiple machines, we won't be able to right? Is there a way we can keep track of the surge within this function itself, a short circuit at the top of the loop maybe?

Or do we not do this because of deletes? Perhaps we should handle this in two phases? A delete phase, then an create phase? Or vice versa?

Given the cap of Machines in the control plane should be about 5 instances, I'm not concerned about the cost of iterating over a loop a few times if it makes the code clearer

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +139,85 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+	}
+
+	// Reconcile any index with no Ready Machines but a replacement pending.
+	for idx, machines := range sortedIndexedMs {
+		if empty(readyMachines(machines)) && any(pendingMachines(machines)) {
+			// There are No Ready Machines for this index but a Pending Machine Replacement is present.
+			// Wait for it to become Ready.
+			logger = logger.WithValues("index", idx, "namespace", r.Namespace, "name", getMachineReplacementName(idx))
+			logger.V(2).Info(waitingForReady)
+			return ctrl.Result{}, nil
```
JoelSpeed, line 174: Should we process all indexes here? Rather than just one?

pkg/controllers/controlplanemachineset/updates.go
damdo: Added some extra explanation 👍

pkg/controllers/controlplanemachineset/updates.go
damdo: Oops, this is a refuse. Yes just pushed the right one.

pkg/controllers/controlplanemachineset/updates.go
damdo: Hhm yeah I can see the functional side of me being tempted by this.
Although it would be purely cosmetic and adds quite a bit of boilerplate to avoid looping over a list a couple of times.

Personally I would keep 1/2 row `for` repetitions as I wouldn't expect loops changing that much and needing a lot of refactoring to justify the extra complexity. But I do not feel strongly about this.

pkg/controllers/controlplanemachineset/updates.go
damdo: > Handling the surge inside another function means that, in the future, if we need to create multiple machines, we won't be able to right?

If in the future we were to increase `maxSurge` to let's say `2`, the current logic would still allow creating multiple machines up to when we hit the surge. It won't allow creating multiple in the same reconciliation though and this is a conscious design choice I made in the code after experimenting things a bit.

Writing the logic in a way that accounts multiple indexes involved at every loop, would complicate the logic significantly.

Unless there already are assumptions on how we do this, I'd prefer to limit reconciliation to a single index per reconcile. This allows us to perform operations in a per index incremental lock step fashion and to keep complexity fairly low.

pkg/controllers/controlplanemachineset/updates.go
damdo: It would break tests as they are written now.
In this case, for example, tests expect just a single "waitingForReady" log before the return.

So for the motivations described here https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56#discussion_r913734444 I'd limit it to one index at a time.

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +138,97 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+
+	// Reconcile any index with no Machine first.
+	for idx, machines := range sortedIndexedMs {
+		if isEmpty(machines) {
+			// There are No Machines for this index.
+			// Create a new Machine for it.
+			logger = logger.WithValues("index", idx, "namespace", r.Namespace, "name", "<Unknown>")
+			return createMachine(ctx, logger, machineProvider, idx, maxSurge, &surgeCount)
+		}
+	}
```
JoelSpeed, line 166: What happens if there's no capacity to surge and an index is empty? 

Eg Index 1 has 2 instances - 1 old, 1 updated, index 0 has 2 instances, 1 old, 1 updated, index 2 has no instances, then we will never get to deleting Machines, instead we will loop on this and fail to create forever

This goes back to, I think we need to do this in multiple phases, do a delete phase, then a create phase

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +138,97 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+
+	// Reconcile any index with no Ready Machines but a replacement pending.
+	for idx, machines := range sortedIndexedMs {
```
JoelSpeed, line 169: Likewise here I think we can do the delete phase first

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +241,124 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+	return currentReplicas - desiredReplicas
+}
+
+// hasAny checks if a MachineInfo slice contains at least 1 element.
+func hasAny(machinesInfo []machineproviders.MachineInfo) bool {
+	return len(machinesInfo) > 0
+}
+
+// any checks if a MachineInfo slice is empty.
+func empty(machinesInfo []machineproviders.MachineInfo) bool {
```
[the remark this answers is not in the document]
damdo: Yes that's in a followup commit already :)

pkg/controllers/controlplanemachineset/updates.go
damdo: Agree, delete phase is now first.

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +241,124 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+	for i := range machinesInfo {
+		if machinesInfo[i].Ready {
+			result = append(result, machinesInfo[i])
+		}
+	}
+	return result
+}
+
+// sortMachineInfos returns a list numerically sorted by index, of each index' MachineInfos.
+func sortMachineInfos(indexedMachineInfos map[int32][]machineproviders.MachineInfo) [][]machineproviders.MachineInfo {
```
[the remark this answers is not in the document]
damdo: Definitely! Added.

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +202,243 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+			logger.V(2).Info(waitingForReady)
+
+			return true, ctrl.Result{}, nil
+		}
+	}
+
+	return false, ctrl.Result{}, nil
+}
+
+func (r *ControlPlaneMachineSetReconciler) reconcileIndexWithOutdatedButUpdated(ctx context.Context, logger logr.Logger, sortedIndexedMs [][]machineproviders.MachineInfo, machineProvider machineproviders.MachineProvider) (bool, ctrl.Result, error) {
```
JoelSpeed, line 361: The name of this function, makes it hard to know what it is actually doing IMO.

I would expect this to be something more like, `deleteReplacedMachines` or something similar, as that is what we are doing, we are looking for any machine that has successfully been replaced, and deleting it

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +202,243 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+		return ctrl.Result{}, werr
+	}
+
+	logger.V(2).Info(createdReplacement)
+	*surgeCount++
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ControlPlaneMachineSetReconciler) reconcileIndexNonReadyButPending(ctx context.Context, logger logr.Logger, sortedIndexedMs [][]machineproviders.MachineInfo) (bool, ctrl.Result, error) {
```
JoelSpeed, line 342: The name of this doesn't make it obvious to me what it's doing, I would expect it to be more inline with what it's achieving, so maybe `waitForPendingIndex` or something like that? We are waiting for an index that has no ready machines but has a pending machine right?

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +202,243 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+			logger.V(2).Info(waitingForRemoved)
+
+			return true, ctrl.Result{}, nil
+		}
+	}
+
+	return false, ctrl.Result{}, nil
+}
+
+func (r *ControlPlaneMachineSetReconciler) reconcileIndexWithOutdatedButPending(ctx context.Context, logger logr.Logger, sortedIndexedMs [][]machineproviders.MachineInfo, machineProvider machineproviders.MachineProvider) (bool, ctrl.Result, error) {
```
JoelSpeed, line 387: As with the others, lets make the name say what this is doing, `waitForPendingMachines`? Why do we do this separately from the `non ready` index, perhaps they can be combined?

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +202,243 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+			logger := logger.WithValues("index", int(outdatedMachine.Index), "namespace", r.Namespace, "name", outdatedMachine.MachineRef.ObjectMeta.Name)
+
+			result, err := createMachine(ctx, logger, machineProvider, outdatedMachine.Index, maxSurge, surgeCount)
+
+			return true, result, err
+		}
+	}
+
+	return false, ctrl.Result{}, nil
+}
```
JoelSpeed, line 444: Perhaps we want to combine these to a `createReplacementMachines` which has two loops, first a not about updating indexes with no machines, then the second loop for those that have a Machine needing an update

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +142,51 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+
+	// Reconcile any index with No Machines.
+	if done, result, err := r.reconcileIndexWithNoMachine(ctx, logger, sortedIndexedMs, machineProvider, maxSurge, &surgeCount); err != nil || done {
+		return result, err
+	}
+
+	// Reconcile any index with any remaining Outdated Machine.
+	if done, result, err := r.reconcileIndexWithOutdated(ctx, logger, sortedIndexedMs, machineProvider, maxSurge, &surgeCount); err != nil || done {
+		return result, err
+	}
```
JoelSpeed, line 185: Because we are using the terms of the classes of Machines, and not really talking about the actions that we are taking, I'm finding the logic of this function hard to follow.

I would have expected something more like a story

```
// Delete machines that are no longer needed

// Wait for any pending machines

// Create any machines that need creating

```

At the moment, if there is a pending Machine, this blocks any further creation right? This would mean, when surge is not 1, we can't actually surge over 1 if a Machine is pending.

Likewise, if there are two Machines in an index that both need an update, but one is pending, we wouldn't be able to continue. Eg, what happens if I update my CPMS with an invalid spec, new machine is pending forever, then I update the spec to fix it. I would expect the CPMS to delete the pending Machine and create a new Machine from the new spec. I don't think the logic right now allows for that

pkg/controllers/controlplanemachineset/updates.go
damdo: That's a good idea 👍

pkg/controllers/controlplanemachineset/updates.go
damdo: Yes these names where mainly a placeholder as I wanted to chat with others to see what they thought would've been good names for them (naming is hard). I like your suggestions for this and the other functions, I'll adopt them 👍

pkg/controllers/controlplanemachineset/updates.go
damdo: See https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56#discussion_r923403427

pkg/controllers/controlplanemachineset/updates.go
damdo: See https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56#discussion_r923403427

pkg/controllers/controlplanemachineset/updates.go
damdo: See https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56#discussion_r923403427

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -55,6 +56,10 @@ const (
 	// the current set of Machines.
 	noUpdatesRequired = "No updates required"
 
+	// noRoomForReplacement is a log message used to inform the user that no room for creating
+	// a replacement Machine is left as maxSurge has been reached.
+	noRoomForReplacement = "No room for creating a replacement Machine, maxSurge reached"
```
JoelSpeed, line 61: I think capacity may be a better word to use here. Perhaps we want to say something like `Insufficient capacity for expansion. Will not create replacement Machine.`?

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -152,3 +195,235 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
...
+		if machinesInfo[i].Ready {
+			result = append(result, machinesInfo[i])
+		}
+	}
+
+	return result
+}
+
+// sortMachineInfosByIndex returns a list numerically sorted by index, of each index' MachineInfos.
+func sortMachineInfosByIndex(indexedMachineInfos map[int32][]machineproviders.MachineInfo) [][]machineproviders.MachineInfo {
```
JoelSpeed, line 257: This function would be good to have some individual unit tests if we don't already have them

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +142,44 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+	// Devise the existing surge and keep track of the current surge count.
+	// No check for early stoppage is done here,
+	// as deletions can continue even if the maxSurge has been already reached.
+	surgeCount := deviseExistingSurge(cpms, sortedIndexedMs)
+
+	// Delete any machines that have successfully been replaced.
+	// This comes first so that we make room to create additional Machines later,
+	// otherwise a failure to update (perhaps because of surge capacity) may prevent deletes,
+	// in turn creating a deadlock scenario.
+	if done, result, err := r.deleteReplacedMachines(ctx, logger, sortedIndexedMs, machineProvider); err != nil || done {
```
JoelSpeed, line 166: Would like to discuss why we return early here if we delete a Machine, not sure that's necessary is it?

JoelSpeed (commented):
Something we additionally need to consider here is what happens when you update the spec mid update, what happens if an index contains two machines, both in need of an update 🤔

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -1274,3 +1388,54 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+var _ = Describe("utils tests", func() {
+	machineGVR := machinev1beta1.GroupVersion.WithResource("machines")
+	nodeGVR := corev1.SchemeGroupVersion.WithResource("nodes")
+	updatedMachineBuilder := resourcebuilder.MachineInfo().
+		WithMachineGVR(machineGVR).
+		WithNodeGVR(nodeGVR).
+		WithReady(true).
+		WithNeedsUpdate(false)
+
+	DescribeTable("should convert an indexed map of MachineInfo into a slice sorted by index",
```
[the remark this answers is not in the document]
damdo: Shuffled around in https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56/commits/f222e14e0d54acf9b79395d9ea94a35affe0b23a

pkg/controllers/controlplanemachineset/updates.go
damdo: We verbally agreed on `Insufficient capacity for expansion, maxSurge is currently being reached. Cannot create a replacement Machine at this time.`

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -137,6 +142,44 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
...
+		return result, err
+	}
+
+	// Wait for Pending Machine Replacements to become ready, if any exist.
+	if done, result, err := r.waitForPendingMachines(ctx, logger, sortedIndexedMs, machineProvider); err != nil || done {
+		return result, err
+	}
```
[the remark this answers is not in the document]
damdo: We talked about this offline, and carried on with the refactoring which allows multiple index and multiple action per index, in each reconcile loop.
See: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56/commits/fc2d4fe1ead847c53876127942b270cea12e128f and: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56/commits/574384e49043514e15f31614ab07bbc390393b72

pkg/controllers/controlplanemachineset/updates.go
damdo: See: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56#discussion_r926855940

pkg/controllers/controlplanemachineset/updates.go
damdo: See: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56#discussion_r926855940

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -497,45 +545,320 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+					},
+					1: {
+						updatedMachineBuilder.WithIndex(1).WithMachineName("machine-1").WithNodeName("node-1").WithNeedsUpdate(true).Build(),
+						updatedMachineBuilder.WithIndex(1).WithMachineName("machine-replacement-1").WithNodeName("node-replacement-1").Build(),
+					},
+					2: {},
+				},
+				setupMock: func() {
+					mockMachineProvider.EXPECT().CreateMachine(gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
+					mockMachineProvider.EXPECT().DeleteMachine(gomock.Any(), gomock.Any(), gomock.Any()).Times(1)
```
JoelSpeed, line 686: Which machine is being deleted?

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -497,45 +545,320 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+								"index", 2,
+								"namespace", namespaceName,
+								"name", "<Unknown>",
+							},
+							Message: noCapacityForExpansion,
+						},
+					}
+				},
+			}),
+			Entry("with an extra index, hitting maxSurge", rollingUpdateTableInput{
```
JoelSpeed, line 723: If there's an extra index, I would the CPMS to remove the additional Machine

JoelSpeed (commented):
This is looking nearly ready to go I think, couple of qs on the tests

pkg/controllers/controlplanemachineset/updates_test.go
JoelSpeed: Discussed with Dam synchronously and we believe this case is valid, we will handle additional indexes in an earlier part of the code

pkg/controllers/controlplanemachineset/updates_test.go
damdo: Pinned in https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/56/commits/dd21fa703f3265703dd24087ce003ff995f640de

PR #57, opened 2022-07-07
Add ownerrefs to machines owned by CPMS

alexander-demichev:
Add ownerrefs to machines owned by CPMS. From function TODO: 
```
TODO: Iterate over the MachineInfos, for each Machine, check the owner references for an owner reference matching
that of the current CPMS (it should be the controller so should be easy to find).
If required, look up the GVK using the rest mapper from the GVR, then uses metav1.PartialObjectMetadata to Patch
the owner references. This should mean we can update the metadata of any type given we know the GVR and existing
ObjectMeta.
```

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -251,48 +251,51 @@ func (r *ControlPlaneMachineSetReconciler) ensureOwnerReferences(ctx context.Con
 			mObjectMeta := mInfo.MachineRef.ObjectMeta
 			mLogger := logger.WithValues("machineNamespace", mObjectMeta.GetNamespace(), "machineName", mObjectMeta.GetName())
 
-			machineGvk, err := r.RESTMapper.KindFor(mInfo.MachineRef.GroupVersionResource)
+			machineGVK, err := r.RESTMapper.KindFor(mInfo.MachineRef.GroupVersionResource)
 			if err != nil {
 				return fmt.Errorf("error getting GVK for machine: %w", err)
 			}
 
 			machine := &metav1.PartialObjectMetadata{}
-			machine.SetGroupVersionKind(machineGvk)
+			machine.SetGroupVersionKind(machineGVK)
 			machine.ObjectMeta = mObjectMeta
 
 			if isOwnedByCurrentCPMS(cpms, machine) {
 				mLogger.V(4).Info("Owner reference already present on machine")
+
 				continue
 			}
 
 			patchBase := client.MergeFrom(machine.DeepCopy())
 
-			err = controllerutil.SetControllerReference(cpms, machine, r.Scheme)
-			if err != nil {
+			if err := controllerutil.SetControllerReference(cpms, machine, r.Scheme); err != nil {
 				mLogger.Error(err, "Cannot add owner reference to machine")
+
 				var alreadyOwnedErr *controllerutil.AlreadyOwnedError
 				if errors.As(err, &alreadyOwnedErr) {
 					meta.SetStatusCondition(&cpms.Status.Conditions, metav1.Condition{
 						Type:               conditionDegraded,
 						Status:             metav1.ConditionTrue,
 						Reason:             reasonMachinesAlreadyOwned,
-						ObservedGeneration: 2,
+						ObservedGeneration: cpms.Generation,
 						Message:            "Observed already owned machine(s) in target machines",
 					})
 
 					// don't return an error here and continue iterating through the machines
 					continue
 				}
+
 				return fmt.Errorf("error setting owner reference: %w", err)
 			}
 
 			if err := r.Client.Patch(ctx, machine, patchBase); err != nil {
-				return fmt.Errorf("error patching the macine: %w", err)
+				return fmt.Errorf("error patching machine: %w", err)
 			}
 
 			mLogger.V(2).Info("Added owner reference to machine")
 		}
 	}
+
 	return nil
 }
 
@@ -333,6 +336,7 @@ func isControlPlaneMachineSetDegraded(cpms *machinev1.ControlPlaneMachineSet) bo
 	return false
 }
 
+// isOwnedByCurrentCPMS determines if the given machine is owned by the ControlPlaneMachineSet.
 func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectMeta *metav1.PartialObjectMetadata) bool {
 	if machineObjectMeta.OwnerReferences == nil {
 		return false
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -18,6 +18,7 @@ package controlplanemachineset
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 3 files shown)


elmiko (commented):
mostly makes sense to me, i just have a question about one of the tests

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -240,11 +242,61 @@ func (r *ControlPlaneMachineSetReconciler) ensureFinalizer(ctx context.Context,
...
+			if err != nil {
+				mLogger.Error(err, "Cannot add owner reference to machine")
+
+				var alreadyOwnedErr *controllerutil.AlreadyOwnedError
+				if errors.As(err, &alreadyOwnedErr) {
+					meta.SetStatusCondition(&cpms.Status.Conditions, metav1.Condition{
+						Type:               conditionDegraded,
+						Status:             metav1.ConditionTrue,
+						Reason:             reasonMachinesAlreadyOwned,
+						ObservedGeneration: 2,
```
RadekManak, line 281: This should be `cpms.Generation`

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -240,11 +242,61 @@ func (r *ControlPlaneMachineSetReconciler) ensureFinalizer(ctx context.Context,
...
+	for _, machineInfo := range machineInfos {
+		for _, mInfo := range machineInfo {
+			if mInfo.MachineRef == nil {
+				continue
+			}
+
+			mObjectMeta := mInfo.MachineRef.ObjectMeta
+			mLogger := logger.WithValues("machineNamespace", mObjectMeta.GetNamespace(), "machineName", mObjectMeta.GetName())
+
+			machineGvk, err := r.RESTMapper.KindFor(mInfo.MachineRef.GroupVersionResource)
```
RadekManak, line 254: nit, capitalize the acronym to `machineGVK`

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -565,45 +577,54 @@ var _ = Describe("ensureOwnerRefrences", func() {
...
+				},
+				ObjectMeta: machineInfos[0][0].MachineRef.ObjectMeta,
+			}
 			machine.ObjectMeta.OwnerReferences = []metav1.OwnerReference{badOwnerReference}
 
-			setControllerErr := controllerutil.SetControllerReference(cpms, machine, testScheme)
-			Expect(setControllerErr).To(HaveOccurred())
-			expectedError = fmt.Errorf("error setting owner reference: %w", setControllerErr)
+			expectedError = controllerutil.SetControllerReference(cpms, machine, testScheme)
+			Expect(expectedError).To(HaveOccurred())
```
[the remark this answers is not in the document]
alexander-demichev: We are building an expected error using a fake machine that should be returned by the controller runtime, this error is later compared with what was really returned, you can see it on line 636 where we test log entries.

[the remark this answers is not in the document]
elmiko: makes sense, thanks for the explanation

elmiko (commented):
/lgtm

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -240,11 +242,61 @@ func (r *ControlPlaneMachineSetReconciler) ensureFinalizer(ctx context.Context,
...
+			machineGVK, err := r.RESTMapper.KindFor(mInfo.MachineRef.GroupVersionResource)
+			if err != nil {
+				return fmt.Errorf("error getting GVK for machine: %w", err)
+			}
+
+			machine := &metav1.PartialObjectMetadata{}
+			machine.SetGroupVersionKind(machineGVK)
+			machine.ObjectMeta = mObjectMeta
+
+			if isOwnedByCurrentCPMS(cpms, machine) {
```
lobziik, line 263: this check already presented in `controllerutil.SetControllerReference`, is this really needed?

JoelSpeed: This allows us to add the debug log, so yes, it's required

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -240,11 +242,61 @@ func (r *ControlPlaneMachineSetReconciler) ensureFinalizer(ctx context.Context,
...
+				mLogger.V(4).Info("Owner reference already present on machine")
+
+				continue
+			}
+
+			patchBase := client.MergeFrom(machine.DeepCopy())
+
+			err = controllerutil.SetControllerReference(cpms, machine, r.Scheme)
+			if err != nil {
```
JoelSpeed, line 272: Scope the error to the if block to avoid shadowing
```suggestion
			if err: = controllerutil.SetControllerReference(cpms, machine, r.Scheme); err != nil {
```

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -459,31 +469,35 @@ var _ = Describe("ensureOwnerRefrences", func() {
 	Context("when some machines already have an existing owner reference", func() {
 		BeforeEach(func() {
 			By("Adding an owner reference to some of the machine infos")
-			skippedOne := false
 			Expect(machineInfos).To(HaveLen(3))
+			Expect(machines).To(HaveLen(3))
+
 			for i := range machineInfos {
-				if !skippedOne {
-					skippedOne = true
```
JoelSpeed, line 466: Did this not work?

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -459,31 +469,35 @@ var _ = Describe("ensureOwnerRefrences", func() {
...
 				}
 
 				for j := range machineInfos[i] {
 					Expect(machineInfos[i][j].MachineRef).ToNot(BeNil())
 					machineInfos[i][j].MachineRef.ObjectMeta.OwnerReferences = []metav1.OwnerReference{expectedOwnerReference}
 				}
+
+				patchBase := client.MergeFrom(machines[i].DeepCopy())
+				machines[i].SetOwnerReferences([]metav1.OwnerReference{expectedOwnerReference})
+				Expect(k8sClient.Patch(ctx, machines[i], patchBase)).To(Succeed())
```
JoelSpeed, line 487: Why do we need to do this? Is the function looking up machines against the API server? Or is this to do with generation and making sure it has the latest machines?

pkg/controllers/controlplanemachineset/controller_test.go
alexander-demichev: I don't where the bug is but it didn't always skip the first machine, sometimes the second or third machine was the one that was skipped.

pkg/controllers/controlplanemachineset/controller_test.go
alexander-demichev: I don't exactly remember but the test was flaking if no reference was present on the server side.

pkg/controllers/controlplanemachineset/controller_test.go
alexander-demichev: oh, right! I didn't notice it a map

pkg/controllers/controlplanemachineset/controller_test.go
JoelSpeed: Ahh yes, it's because we check the owner references are present on the server, but we haven't set them on the server, good catch

JoelSpeed (commented):
Small nit otherwise LGTM

JoelSpeed:
Once you fixed the nit, please squash the fixups into the feature commit to make the history clean

PR #58, opened 2022-07-07
[OCPCLOUD-1617] Implement OpenShift Machine Provider Interface for CPMS

Fedosin:
The Machine Provider interface abstracts the machine API specific information out of the core of the CPMS operator.

This commit implements the interface that allows to gather machine information, create and delete machines.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -24,6 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders"
@@ -139,5 +140,53 @@ func (m *openshiftMachineProvider) CreateMachine(ctx context.Context, logger log
 
 // DeleteMachine deletes the Machine references in the machineRef provided.
 func (m *openshiftMachineProvider) DeleteMachine(ctx context.Context, logger logr.Logger, machineRef *machineproviders.ObjectRef) error {
+	machinesGVR := machinev1beta1.GroupVersion.WithResource("machines")
+
+	if machineRef.GroupVersionResource != machinesGVR {
+		logger.Error(errUnknownGroupVersionResource,
+			"Could not delete machine",
+			"expectedGVR", machinesGVR.String(),
+			"gotGVR", machineRef.GroupVersionResource.String(),
+		)
+
+		return fmt.Errorf("%w: expected %s, got %s", errUnknownGroupVersionResource, machinesGVR.String(), machineRef.GroupVersionResource.String())
+	}
+
+	machine := machinev1beta1.Machine{
+		ObjectMeta: machineRef.ObjectMeta,
+	}
+
+	if err := m.client.Delete(ctx, &machine); err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.V(2).Info(
+				"Machine not found",
+				"namespace", machineRef.ObjectMeta.Namespace,
+				"machineName", machineRef.ObjectMeta.Name,
+				"group", machinev1beta1.GroupVersion.Group,
+				"version", machinev1beta1.GroupVersion.Version,
+			)
+
+			return nil
+		}
+
+		logger.Error(err,
+			"Could not delete machine",
+			"namespace", machineRef.ObjectMeta.Namespace,
+			"machineName", machineRef.ObjectMeta.Name,
+			"group", machinev1beta1.GroupVersion.Group,
+			"version", machinev1beta1.GroupVersion.Version,
+		)
+
+		return fmt.Errorf("could not delete machine %s in namespace %s: %w", machineRef.ObjectMeta.Name, machineRef.ObjectMeta.Namespace, err)
+	}
+
+	logger.V(2).Info(
+		"Deleted machine",
+		"namespace", machineRef.ObjectMeta.Namespace,
+		"machineName", machineRef.ObjectMeta.Name,
+		"group", machinev1beta1.GroupVersion.Group,
+		"version", machinev1beta1.GroupVersion.Version,
+	)
+
 	return nil
 }
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -26,12 +26,21 @@ import (
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachineryruntime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/rand"
 
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+const (
+	// openshiftMachineRoleLabel is the OpenShift Machine API machine role label.
+	// This must be present on all OpenShift Machine API Machine templates.
+	openshiftMachineRoleLabel = "machine.openshift.io/cluster-api-machine-role"
 )
 
 var (
@@ -49,6 +58,11 @@ var (
 	// is not present and therefore a Machine cannot be created. The Cluster ID is required to construct the name for
 	// the new Machines.
 	errMissingClusterIDLabel = fmt.Errorf("missing required label on machine template metadata: %s", machinev1beta1.MachineClusterIDLabel)
+
+	// errMissingMachineRoleLabel is used to denote that the machine role label, expected to be on the Machine template
+	// is not present and therefore a Machine cannot be created. The Machine Role is required to construct the name for
+	// the new Machines.
+	errMissingMachineRoleLabel = fmt.Errorf("missing required label on machine template metadata: %s", openshiftMachineRoleLabel)
 
 	// errUnexpectedMachineType is used to denote that the machine provider was requested
 	// for an unsupported machine provider type (ie not OpenShift Machine v1beta1).
@@ -84,6 +98,15 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
 		return nil, fmt.Errorf("error mapping machine indexes: %w", err)
 	}
 
+	machineAPIScheme := apimachineryruntime.NewScheme()
+	if err := machinev1.Install(machineAPIScheme); err != nil {
+		return nil, fmt.Errorf("unable to add machine.openshift.io/v1 scheme: %w", err)
+	}
+
+	if err := machinev1beta1.Install(machineAPIScheme); err != nil {
+		return nil, fmt.Errorf("unable to add machine.openshift.io/v1beta1 scheme: %w", err)
+	}
+
 	return &openshiftMachineProvider{
 		client:               cl,
 		indexToFailureDomain: indexToFailureDomain,
@@ -91,6 +114,8 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
 		machineTemplate:      *cpms.Spec.Template.OpenShiftMachineV1Beta1Machine,
 		ownerMetadata:        cpms.ObjectMeta,
 		providerConfig:       providerConfig,
+		namespace:            cpms.Namespace,
+		machineAPIScheme:     machineAPIScheme,
 	}, nil
 }
 
@@ -118,6 +143,12 @@ type openshiftMachineProvider struct {
 
 	// providerConfig stores the providerConfig for creating new Machines.
 	providerConfig providerconfig.ProviderConfig
+
+	// namespace store the namespace where new machines will be created.
+	namespace string
+
+	// machineAPIScheme contains scheme for Machine API v1 and v1beta1.
+	machineAPIScheme *apimachineryruntime.Scheme
 }
 
 // GetMachineInfos inspects the current state of the Machines matched by the selector
@@ -135,7 +166,76 @@ func (m *openshiftMachineProvider) GetMachineInfos(ctx context.Context, logger l
 // CreateMachine creates a new Machine from the template provider config based on the
 // failure domain index provided.
 func (m *openshiftMachineProvider) CreateMachine(ctx context.Context, logger logr.Logger, index int32) error {
+	machineName, err := m.getMachineName(index)
+	if err != nil {
+		return fmt.Errorf("could not generate machine name: %w", err)
+	}
+
+	cpms := &machinev1.ControlPlaneMachineSet{
+		ObjectMeta: m.ownerMetadata,
+	}
+
+	machine := &machinev1beta1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        machineName,
+			Namespace:   m.namespace,
+			Annotations: m.machineTemplate.ObjectMeta.Annotations,
+			Labels:      m.machineTemplate.ObjectMeta.Labels,
+		},
+		Spec: m.machineTemplate.Spec,
+	}
+
+	pc, err := m.providerConfig.InjectFailureDomain(m.indexToFailureDomain[index])
+	if err != nil {
+		return fmt.Errorf("cannot inject failure domain in the provider config: %w", err)
+	}
+
+	rawConfig, err := pc.RawConfig()
+	if err != nil {
+		return fmt.Errorf("cannot fetch raw config from provider config: %w", err)
+	}
+
+	machine.Spec.ProviderSpec.Value.Raw = rawConfig
+
+	if err := controllerutil.SetControllerReference(cpms, machine, m.machineAPIScheme); err != nil {
+		return fmt.Errorf("could not set owner reference: %w", err)
+	}
+
+	if err := m.client.Create(ctx, machine); err != nil {
+		logger.Error(err,
+			"Could not create machine",
+			"namespace", machine.ObjectMeta.Namespace,
+			"machineName", machine.ObjectMeta.Name,
+			"group", machinev1beta1.GroupVersion.Group,
+			"version", machinev1beta1.GroupVersion.Version,
+		)
+
+		return fmt.Errorf("cannot create machine %s in namespace %s: %w", machine.Name, machine.Namespace, err)
+	}
+
+	logger.V(2).Info(
+		"Created machine",
+		"index", index,
+		"machineName", machine.Name,
+		"failureDomain", m.indexToFailureDomain[index].String(),
+	)
+
 	return nil
+}
+
+// getMachineName generates a machine name based on the index.
+func (m *openshiftMachineProvider) getMachineName(index int32) (string, error) {
+	clusterID, ok := m.machineTemplate.ObjectMeta.Labels[machinev1beta1.MachineClusterIDLabel]
+	if !ok {
+		return "", errMissingClusterIDLabel
+	}
+
+	machineRole, ok := m.machineTemplate.ObjectMeta.Labels[openshiftMachineRoleLabel]
+	if !ok {
+		return "", errMissingMachineRoleLabel
+	}
+
+	return fmt.Sprintf("%s-%s-%s-%d", clusterID, machineRole, rand.String(5), index), nil
 }
 
 // DeleteMachine deletes the Machine references in the machineRef provided.
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -20,14 +20,18 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 
 	"github.com/go-logr/logr"
 	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+	}
+
+	templateProviderConfig, err := m.providerConfig.InjectFailureDomain(failureDomain)
+	if err != nil {
+		return nil, fmt.Errorf("error injecting failure domain into provider config: %w", err)
+	}
+
+	configsEqual, err := templateProviderConfig.Equal(providerConfig)
+	if err != nil {
+		return nil, fmt.Errorf("cannot compare provider configs: %w", err)
```
elmiko, line 248: this error seems a little cryptic, is there something that the user could do here to fix things?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+		MachineRef:   machineRef,
+		NodeRef:      nodeRef,
+		Ready:        *machine.Status.Phase == runningPhase,
+		NeedsUpdate:  !configsEqual || !indexFound,
+		Index:        index,
+		ErrorMessage: errorMessage,
+	}, nil
+}
+
+// failureDomainToIndex returns the index of failure domain in the list.
```
elmiko, line 261: what about the boolean it returns?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+	rawConfig, err := pc.RawConfig()
+	if err != nil {
+		return fmt.Errorf("cannot fetch raw config from provider config: %w", err)
+	}
+
+	machine.Spec.ProviderSpec.Value.Raw = rawConfig
+
+	if err := m.client.Create(ctx, &machine); err != nil {
+		logger.Error(err,
+			"Could not delete machine",
```
elmiko, line 360: should this be `"Could not create machine"` ?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+			"namespace", machine.ObjectMeta.Namespace,
+			"machineName", machine.ObjectMeta.Name,
+			"group", machinev1beta1.GroupVersion.Group,
+			"version", machinev1beta1.GroupVersion.Version,
+		)
+
+		return fmt.Errorf("cannot create machine %s: %w", machine.Name, err)
```
elmiko, line 367: i like the error message you added for the delete machine, could we add the namespace here as well?

eg

`"cannot create machine %s in namespace %s: %w"`

elmiko (commented):
mostly looks good to me, just a couple questions/comments

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -59,7 +83,7 @@ var (
 )
 
 // NewMachineProvider creates a new OpenShift Machine v1beta1 machine provider implementation.
-func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Client, cpms *machinev1.ControlPlaneMachineSet) (machineproviders.MachineProvider, error) {
+func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Client, cpms *machinev1.ControlPlaneMachineSet, namespace string) (machineproviders.MachineProvider, error) {
```
JoelSpeed, line 86: The namespace should come from the CPMS, not from whatever is calling this. CPMS objects can only own objects in their own namespace, so it must come from there

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not convert label selector to selector: %w", err)
+	}
+
+	machineList := &machinev1beta1.MachineList{}
+	if err := m.client.List(ctx, machineList, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("failed to list machines: %w", err)
+	}
+
+	correctFormat := isMachinesNamesFormatCorrect(machineList.Items)
```
JoelSpeed, line 171: If you are doing this, you need to do it on a per machine basis, you can't assume that all Machines will meet the format or not. If the Machines don't meet the format when the CPMS first starts, but we then create some more machines, the new Machines will match the format, and you'll have a mix

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+
+	failureDomains, err := providerconfig.ExtractFailureDomainsFromMachines(machineList.Items)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract failure domains from machines: %w", err)
+	}
+
+	for i, machine := range machineList.Items {
+		machineInfo, err := m.generateMachineInfo(machine, failureDomains[i], correctFormat, logger)
```
JoelSpeed, line 179: Why do we pass the failure domains in here? The Machine info should come from the Machine itself, I wouldn't expect us to be comparing a randomly ordered list of machines with a randomly ordered list of failure domains and trying to match those, this is what the mapping logic is for right?

This should be done on a machine by machine basis else it becomes a bit disjoint

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+
+// generateMachineInfo creates a MachineInfo object for a given machine.
+func (m *openshiftMachineProvider) generateMachineInfo(machine machinev1beta1.Machine, failureDomain failuredomain.FailureDomain, correctFormat bool, logger logr.Logger) (*machineproviders.MachineInfo, error) {
+	machineRef := getMachineRef(machine)
+
+	nodeRef := getNodeRef(machine)
+
+	errorMessage := getMachineErrorMessage(machine)
+
+	index, indexFound := m.failureDomainToIndex(failureDomain)
```
JoelSpeed, line 213: You should extract the failure domain from the provider config on the Machine, not have it passed in I think

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+
+	return *machine.Status.ErrorMessage
 }
 
 // CreateMachine creates a new Machine from the template provider config based on the
 // failure domain index provided.
 func (m *openshiftMachineProvider) CreateMachine(ctx context.Context, logger logr.Logger, index int32) error {
+	machineName, err := m.getMachineName(index)
+	if err != nil {
+		return err
```
JoelSpeed, line 321:
```suggestion
		return fmt.Errorf("could not generate machine name: %w", err)
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+	ownerReferences := []metav1.OwnerReference{
+		{
+			APIVersion:         machinev1.GroupVersion.String(),
+			Kind:               "ControlPlaneMachineSet",
+			Name:               m.ownerMetadata.Name,
+			UID:                m.ownerMetadata.UID,
+			Controller:         pointer.Bool(true),
+			BlockOwnerDeletion: pointer.Bool(true),
+		},
+	}
```
JoelSpeed, line 333: Use the controllerutil utility for this, I think it's `SetControllerRef`

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+	if !ok {
+		return "", errMissingMachineRoleLabel
+	}
+
+	return fmt.Sprintf("%s-%s-%s-%d", clusterID, machineRole, rand.String(5), index), nil
+}
+
 // DeleteMachine deletes the Machine references in the machineRef provided.
 func (m *openshiftMachineProvider) DeleteMachine(ctx context.Context, logger logr.Logger, machineRef *machineproviders.ObjectRef) error {
+	if machineRef.GroupVersionResource != machinev1beta1.GroupVersion.WithResource("machines") {
```
JoelSpeed, line 397: As you use this a couple of times, why not make a var from it
```suggestion
	machinesGVR := machinev1beta1.GroupVersion.WithResource("machines")

	if machineRef.GroupVersionResource != machinesGVR {
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +156,289 @@ type openshiftMachineProvider struct {
...
+		"machineName", machine.Name,
+		"failureDomain", m.indexToFailureDomain[index].String(),
+	)
+
 	return nil
 }
 
+// getMachineName generates a machine name based on the index.
+func (m *openshiftMachineProvider) getMachineName(index int32) (string, error) {
+	clusterID, ok := m.machineTemplate.ObjectMeta.Labels[machinev1beta1.MachineClusterIDLabel]
```
JoelSpeed, line 382: Is the cluster ID label required on the Machine? Or is it just required on the CPMS? Can we double check this, might be we need to look it up from the CPMS instead

pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
```diff
@@ -64,20 +73,22 @@ var _ = Describe("MachineProvider", func() {
 	})
 
 	Context("GetMachineInfos", func() {
-		const clusterID = "cpms-machine-info-test-id"
+		const clusterID = "cpms-cluster-test-id"
```
JoelSpeed, line 76: Why not use a clusterID specific to this test case?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
```diff
@@ -43,10 +43,19 @@ import (
 var _ = Describe("MachineProvider", func() {
 	const ownerUID = "uid-1234abcd"
 	const ownerName = "machineOwner"
+	const awsSubnetName = "aws-subnet-12345678"
```
JoelSpeed, line 46: Can we make sure to use different subnets for different availability zones please

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -83,13 +108,24 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
...
 
+	machineAPIScheme := apimachineryruntime.NewScheme()
+	if err := machinev1.Install(machineAPIScheme); err != nil {
+		return nil, fmt.Errorf("unable to add machine.openshift.io/v1 scheme: %w", err)
+	}
+
+	if err := machinev1beta1.Install(machineAPIScheme); err != nil {
+		return nil, fmt.Errorf("unable to add machine.openshift.io/v1beta1 scheme: %w", err)
+	}
+
```
JoelSpeed, line 119: Is it possible to pass the scheme through from the main controller? It should have the scheme initialised already right? 

Or is the plan here to use a more succinct scheme because we can?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -117,6 +153,12 @@ type openshiftMachineProvider struct {
 
 	// providerConfig stores the providerConfig for creating new Machines.
 	providerConfig providerconfig.ProviderConfig
+
+	// namespace store the namespace where new machines will be created
+	namespace string
+
+	// machineAPIScheme contains scheme for Machine API v1 and v1beta1
+	machineAPIScheme *apimachineryruntime.Scheme
```
JoelSpeed, line 161: Nit, comments should end in periods
```suggestion
	// namespace store the namespace where new machines will be created.
	namespace string

	// machineAPIScheme contains scheme for Machine API v1 and v1beta1.
	machineAPIScheme *apimachineryruntime.Scheme
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: This comment needs checking

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+	if err := m.client.List(ctx, machineList, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("failed to list machines: %w", err)
+	}
+
+	correctFormat := isMachinesNamesFormatCorrect(machineList.Items)
+
+	for _, machine := range machineList.Items {
+		machineInfo, err := m.generateMachineInfo(machine, correctFormat, logger)
+		if err != nil {
+			return nil, err
```
JoelSpeed, line 190: This error should be wrapped with more info
```suggestion
			return nil, fmt.Errorf("could not generate machine info for machine %s: %w", machine.Name, err)
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+		}
+
+		machineInfos = append(machineInfos, *machineInfo)
```
JoelSpeed, line 193: Why do we have to dereference here? I wouldn't expect the generateMachineInfo function to return a pointer if we don't need to

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+		nodeName := ""
+		if machineInfo.NodeRef != nil {
+			nodeName = machineInfo.NodeRef.ObjectMeta.Name
+		}
+
+		logger.V(4).Info(
+			"Gathered Machine Info",
+			"machineName", machine.GetName(),
+			"nodeName", nodeName,
+			"index", int(machineInfo.Index),
```
JoelSpeed, line 204: If this conversion is to satisfy the tests, we should update the tests I think

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+			"needsUpdate", machineInfo.NeedsUpdate,
+			"errorMessage", machineInfo.ErrorMessage,
+		)
+	}
+
+	return machineInfos, nil
+}
+
+// generateMachineInfo creates a MachineInfo object for a given machine.
+func (m *openshiftMachineProvider) generateMachineInfo(machine machinev1beta1.Machine, correctFormat bool, logger logr.Logger) (*machineproviders.MachineInfo, error) {
```
JoelSpeed, line 215: Why return a pointer?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+
+	failureDomain, err := providerconfig.ExtractFailureDomainFromMachine(machine)
+	if err != nil {
+		return nil, fmt.Errorf("cannot extract failure domain from machine: %w", err)
+	}
+
+	index, indexFound := m.failureDomainToIndex(failureDomain)
+	if !indexFound && !correctFormat {
+		// When the machine names do not fit the pattern, and the failure domains are
+		// not recognised, returns an error.
```
JoelSpeed, line 232: Is it possible to get to this case? How would a user resolve this issue and is there a way we can resolve it for them?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+		logger.Error(errCouldNotDetermineMachineIndex,
+			"Could not gather Machine Info",
+		)
+
+		return nil, errCouldNotDetermineMachineIndex
+	}
+
+	// If machine name index and domain index are different, we should use the machine
+	// name index in the result, but only if the name is correctly formatted
+	if correctFormat {
```
JoelSpeed, line 242: Is this not handled by the reconcile machine mappings logic we already have? Shouldn't it account for this so we don't need to look up from the machine name in this case?

Perhaps this is the logic we need when `!indexFound` rather than returning an error?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: Maybe this will help a little, not sure what exactly the errors coming out may be though
```suggestion
		return nil, fmt.Errorf("could not compare existing and desired provider configs: %w", err)
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -20,20 +20,40 @@ import (
...
+	runningPhase = "Running"
+
+	// openshiftMachineRoleLabel is the OpenShift Machine API machine role label.
+	// This must be present on all OpenShift Machine API Machine templates.
+	openshiftMachineRoleLabel = "machine.openshift.io/cluster-api-machine-role"
 )
 
 var (
+	// digitCheck is a regexp to check that machine name suffixes has up to 3 digits.
+	digitCheck = regexp.MustCompile(`^[\d]{1,3}$`)
```
JoelSpeed, line 55: Up to 3 digits would mean we can have up to 999 control plane machines? Perhaps we want to limit this to 2 digits and 99 control plane machines?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+	}
+}
+
+// getMachineErrorMessage returns error message for the given machine.
+func getMachineErrorMessage(machine machinev1beta1.Machine) string {
+	if machine.Status.ErrorMessage == nil {
+		return ""
+	}
+
+	return *machine.Status.ErrorMessage
```
JoelSpeed, line 330: This could just be `pointer.StringDeref`, `k8s.io/utils/pointer`, perhaps we don't need this func?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +170,296 @@ type openshiftMachineProvider struct {
...
+
+	rawConfig, err := pc.RawConfig()
+	if err != nil {
+		return fmt.Errorf("cannot fetch raw config from provider config: %w", err)
+	}
+
+	machine.Spec.ProviderSpec.Value.Raw = rawConfig
+
+	err = controllerutil.SetControllerReference(cpms, machine, m.machineAPIScheme)
+	if err != nil {
```
JoelSpeed, line 373: Inline this and don't shadow please
```suggestion
	if err = controllerutil.SetControllerReference(cpms, machine, m.machineAPIScheme); err != nil {
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
Fedosin: Well, we can, but we need to overwrite a lot of test cases. Currently we have just one provider config with this particular subnet. If we'd like to make 3 different configs, one for each subnet, we have to modify tests' inputs.

Fedosin: I created an additional commit to fix it. Please tell me if you like how it looks.

pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
Fedosin: I think because it is [hardcoded](https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/test/resourcebuilder/consts.go#L24) in the CPMS description: https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/test/resourcebuilder/control_plane_machine_set.go#L57.

Generally speaking I can make it modifiable, but is there any important reason for this? We don't use custom cluster IDs in other tests/

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
Fedosin: To be able to write `return nil, err`. Now I updated it to return structs.

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
Fedosin: Yeah, that was the idea - to limit the scheme scope.

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: This is an open question still

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +167,274 @@ type openshiftMachineProvider struct {
...
+		)
+
+		return machineproviders.MachineInfo{}, errCouldNotDetermineMachineIndex
+	}
+
+	// If machine name index and domain index are different, we should use the machine
+	// name index in the result, but only if the name is correctly formatted
+	if correctFormat && int32(machineNameIndex) != index {
+		index = int32(machineNameIndex)
+	}
```
JoelSpeed, line 240: Can we add a comment to explain why this might happen? I guess if a failure domain was removed? Or if we scaled down from 5 replicas to 3?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -128,16 +167,274 @@ type openshiftMachineProvider struct {
...
+
+	rawConfig, err := pc.RawConfig()
+	if err != nil {
+		return fmt.Errorf("cannot fetch raw config from provider config: %w", err)
+	}
+
+	machine.Spec.ProviderSpec.Value.Raw = rawConfig
+
+	if err = controllerutil.SetControllerReference(cpms, machine, m.machineAPIScheme); err != nil {
```
JoelSpeed, line 348: Nit, still shadowing
```suggestion
	if err := controllerutil.SetControllerReference(cpms, machine, m.machineAPIScheme); err != nil {
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
JoelSpeed: I think it's just so that we can see that the test cases are working, it's not important

JoelSpeed (commented):
I think this is nearly good to go, couple of little bits left though

PR #61, opened 2022-07-12
[OCPCLOUD-1486] ControlPlaneMachineSet deletion logic

lobziik:
During ControlPlaneMachineSet deletion procedure
reconciler removes owner reference from the machines which it owns.

This pr introduces such logic, related tests and helper functions.

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -228,7 +228,95 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
 // Once the owner references are removed, it removes the finalizer to allow the garbage collector to reap
 // the deleted ControlPlaneMachineSet.
 func (r *ControlPlaneMachineSetReconciler) reconcileDelete(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet) (ctrl.Result, error) {
+	logger.V(1).Info("Reconciling control plane machine set deletion")
+
+	machineTypeMeta, err := providers.GetMachineTypeMeta(cpms.Spec.Template.MachineType)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not determine machines type: %w", err)
+	}
+
+	logger.V(4).Info("Detected machines type", "kind", machineTypeMeta.Kind, "apiVersion", machineTypeMeta.APIVersion)
+
+	selector, err := metav1.LabelSelectorAsSelector(&cpms.Spec.Selector)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not convert label selector to selector: %w", err)
+	}
+
+	machinesMeta := &metav1.PartialObjectMetadataList{
+		TypeMeta: machineTypeMeta,
+	}
+	if err := r.List(ctx, machinesMeta, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list machines: %w", err)
+	}
+
+	var errs []error
+
+	for i := range machinesMeta.Items {
+		machineObjectMeta := machinesMeta.Items[i]
+		logger.V(4).Info("Removing owner reference", "machine", machineObjectMeta.Name)
+		patchBase := client.MergeFrom(machineObjectMeta.DeepCopy())
+
+		if !removeOwnerRef(&machineObjectMeta, cpms) {
+			continue
+		}
+
+		if err := r.Patch(ctx, &machineObjectMeta, patchBase); err != nil {
+			errs = append(errs, fmt.Errorf("error removing owner reference from machine %s: %w", machineObjectMeta.Name, err))
+		}
+	}
+
+	if len(errs) > 0 {
+		return ctrl.Result{}, errorutils.NewAggregate(errs)
+	}
+
+	cpmsCopy := cpms.DeepCopy()
+
+	finalizerUpdated := controllerutil.RemoveFinalizer(cpmsCopy, controlPlaneMachineSetFinalizer)
+	if finalizerUpdated {
+		logger.V(4).Info("Removing finalizer from the control plane machine set")
+
+		if err := r.Update(ctx, cpmsCopy); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update control plane machine set: %w", err)
+		}
+	}
+
 	return ctrl.Result{}, nil
+}
+
+// hasOwnerRef returns true if target has an ownerRef to owner.
+func hasOwnerRef(target, owner client.Object) bool {
+	ownerUID, ownerName := owner.GetUID(), owner.GetName()
+	for _, ownerRef := range target.GetOwnerReferences() {
+		if ownerRef.UID == ownerUID &&
+			ownerRef.Name == ownerName {
+			return true
+		}
+	}
+
+	return false
+}
+
+// removeOwnerRef removes owner from the ownerRefs of target, if necessary. Returns true if target needs
+// to be updated and false otherwise.
+func removeOwnerRef(target, owner client.Object) bool {
+	if !hasOwnerRef(target, owner) {
+		return false
+	}
+
+	ownerUID := owner.GetUID()
+	newRefs := []metav1.OwnerReference{}
+
+	for _, ref := range target.GetOwnerReferences() {
+		if ref.UID == ownerUID {
+			continue
+		}
+
+		newRefs = append(newRefs, ref)
+	}
+
+	target.SetOwnerReferences(newRefs)
+
+	return true
 }
 
 // ensureFinalizer adds a finalizer to the ControlPlaneMachineSet if required.
--- a/pkg/controllers/controlplanemachineset/controller_test.go
+++ b/pkg/controllers/controlplanemachineset/controller_test.go
@@ -177,16 +177,62 @@ var _ = Describe("With a running controller", func() {
 			Expect(k8sClient.Delete(ctx, cpms)).To(Succeed())
 		})
 
-		PIt("should eventually be removed", func() {
+		It("should eventually be removed", func() {
 			Eventually(komega.Get(cpms)).Should(MatchError("controlplanemachinesets.machine.openshift.io \"cluster\" not found"))
 		})
 
-		PIt("should remove the owner references from the Machines", func() {
+		It("should remove the owner references from the Machines", func() {
 			Eventually(komega.ObjectList(&machinev1beta1.MachineList{})).Should(HaveField("Items", SatisfyAll(
 				HaveLen(3),
 				HaveEach(HaveField("ObjectMeta.OwnerReferences", HaveLen(0))),
 			)), "3 Machines should exist, each should have no owner references")
 		})
+	})
+})
+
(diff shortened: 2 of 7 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig.go
```diff
@@ -203,6 +207,11 @@ func getPlatformTypeFromMachineTemplate(tmpl machinev1.OpenShiftMachineV1Beta1Ma
 // getPlatformTypeFromProviderSpec determines machine platform from the providerSpec.
 // The providerSpec object's kind field is unmarshalled and the platform type is inferred from it.
 func getPlatformTypeFromProviderSpec(providerSpec machinev1beta1.ProviderSpec) (configv1.PlatformType, error) {
+	// providerSpec.Value might be nil, return early with error if so
+	if providerSpec.Value == nil {
+		return "", fmt.Errorf("%w: can't determine platform type, provider spec is empty", errProviderSpecIsEmpty)
```
JoelSpeed, line 212: Nit, please don't abbreviate in error messages
```suggestion
		return "", fmt.Errorf("%w: cannot determine platform type, provider spec is empty", errProviderSpecIsEmpty)
```

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -226,9 +228,99 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
...
+	newRefs := make([]metav1.OwnerReference, len(oldRefs)-1)
+	skip := 0
+
+	for i := range oldRefs {
+		if oldRefs[i].UID == ownerUID {
+			skip = -1
+		} else {
+			newRefs[i+skip] = oldRefs[i]
+		}
+	}
```
JoelSpeed, line 317: I'm not really concerned about the efficiency here, we are working with small amounts of data, is this easier to read if it is. Id rather we make the code clean and easy to read/maintain than optimise efficiency at this point
```
newRefs := []metav1.OwnerReference{}

for _, ref := range target.GetOwnerReferences() {
  if ref.UID == ownerUID {
    continue
  }

  append(newRefs, ref)
}
```

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -189,6 +190,52 @@ var _ = Describe("With a running controller", func() {
 	})
 })
 
+var _ = Describe("ownerRef helpers", Ordered, func() {
```
JoelSpeed, line 193: Generally we avoid ordered containers as they mean that side effects of previous tests can influence following tests. What's the motivation behind this set being ordered?

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -226,9 +228,99 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not convert label selector to selector: %w", err)
+	}
+
+	machinesMeta := &metav1.PartialObjectMetadataList{
+		TypeMeta: machinesTypeMeta,
+	}
+	if err := r.List(ctx, machinesMeta, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list machines: %w", err)
+	}
```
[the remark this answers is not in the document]
lobziik: Couple of things which been in my mind, when i made this:

1) Need merge MachineInfos getter first (https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/58)
2) as i understand, this will require provider initialisation first, which require some conditions to be fulfilled (https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go#L67 as example)

I agree, that it would be better to leverage MachineProvider interface, but it's initialization logic should be reconsidered a bit then.

pkg/controllers/controlplanemachineset/controller_test.go
lobziik: Actually, i just wanted to use `BeforeAll` block.  Out of the docs i made a conclusion that this should affect only to this container. However i didnt look inside ginkgo and i'm not sure how it might actually influence following tests outside this container.

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -226,9 +228,99 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
...
+
+		if err := r.Update(ctx, cpmsCopy); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to update control plane machine set: %w", err)
+		}
+	}
+
 	return ctrl.Result{}, nil
 }
 
+// hasOwnerRef returns true if target has an ownerRef to owner.
```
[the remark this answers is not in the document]
lobziik: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil@v0.12.1#pkg-functions Seems not.

pkg/controllers/controlplanemachineset/controller_test.go
JoelSpeed: You're correct in this only affecting this container, but my concern is that as we add more tests, these will be hard to work out whats happening. For example, if 1 test removes and the next re-adds, you have to read through all previous tests to understand the state right? What's the disadvantage of using a `BeforeEach` on this? That you have to clean up?

lobziik: There are a bunch of objects created without any further changes changes. Tests itself are stateless.

>What's the disadvantage of using a BeforeEach on this?
A bunch of unnecessary actions, basically. And necessity of cleaning it up each time then.

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -228,9 +228,97 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
...
 // the deleted ControlPlaneMachineSet.
 func (r *ControlPlaneMachineSetReconciler) reconcileDelete(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet) (ctrl.Result, error) {
+	logger.V(1).Info("Reconciling control plane machine set deletion")
+
+	machinesTypeMeta, err := providers.GetMachinesTypeMeta(cpms.Spec.Template.MachineType)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not determine machines type: %w", err)
+	}
+
+	logger.V(4).Info("Detected machines type", "kind", machinesTypeMeta.Kind, "version", machinesTypeMeta.APIVersion)
```
JoelSpeed, line 238: Kind will always be Machine right, did you want the group here instead?

pkg/controllers/controlplanemachineset/controller.go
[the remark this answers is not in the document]
JoelSpeed: Ack that makes sense, we can keep like this

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -228,9 +228,97 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
...
+	}
+	if err := r.List(ctx, machinesMeta, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to list machines: %w", err)
+	}
+
+	var errs []error
+
+	for i := range machinesMeta.Items {
+		machineObjectMeta := machinesMeta.Items[i]
+		logger.V(4).Info("Try to remove owner reference", "machine", machineObjectMeta.Name)
```
JoelSpeed, line 256: I don't think we need to say try, else you'd expect another message to say we succeeded, let's do this and then log an error if there's an error
```suggestion
		logger.V(4).Info("Removing owner reference", "machine", machineObjectMeta.Name)
```

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -228,9 +228,97 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachines(ctx context.Context
...
+
+	if len(errs) > 0 {
+		return ctrl.Result{}, errorutils.NewAggregate(errs)
+	}
+
+	cpmsCopy := cpms.DeepCopy()
+
+	finalizerUpdated := controllerutil.RemoveFinalizer(cpmsCopy, controlPlaneMachineSetFinalizer)
+	if finalizerUpdated {
+		logger.V(4).Info("Remove finalizer from the control plane machine set")
```
JoelSpeed, line 276: Nit
```suggestion
		logger.V(4).Info("Removing finalizer from the control plane machine set")
```

pkg/machineproviders/providers/machineproviders.go
```diff
@@ -50,3 +52,19 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
 		return nil, fmt.Errorf("%w: %s", errUnexpectedMachineType, cpms.Spec.Template.MachineType)
 	}
 }
+
+// GetMachinesTypeMeta returns proper TypeMeta from ControlPlaneMachineSetMachineType.
+func GetMachinesTypeMeta(cpmsMachinesType machinev1.ControlPlaneMachineSetMachineType) (metav1.TypeMeta, error) {
```
JoelSpeed, line 57: Nit, I'd remove the S here
```suggestion
// GetMachineTypeMeta returns proper TypeMeta from ControlPlaneMachineSetMachineType.
func GetMachineTypeMeta(cpmsMachinesType machinev1.ControlPlaneMachineSetMachineType) (metav1.TypeMeta, error) {
```

pkg/machineproviders/providers/machineproviders.go
```diff
@@ -50,3 +52,19 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
...
+	// TODO perhaps move this to respective provider's packages ?
```
JoelSpeed, line 58: I think we can drop this TODO, this is ok here for now and we can work out if we want to refactor another time

JoelSpeed (commented):
Thanks for responding to my various questions, few nits, but otherwise I think lets merge this logic as is

pkg/controllers/controlplanemachineset/controller.go
JoelSpeed: Ahh right, this is a TypeMeta not a GroupVersionKind 🤦 In which case, to keep terms the same, lets change `"version"` to `"apiVersion"` and then I'm happy to approve

lobziik: I just thought about it a bit, do we have some particular use-case, where we might want to have an extracted group there? Current variant should be fine, if we gonna just use grep against this logs.

JoelSpeed: Yeah I think the current variant is fine, it gives us the information we need to work out which type of machine we are working with, I'd just prefer we use the `apiVersion` term as this is what you see in K8s normally.

APIVersion is different to Version because the APIVersion contains both the Group and Version from the GroupVersionKind trio

JoelSpeed (commented):
/approve
/lgtm

PR #62, opened 2022-07-13
[OCPCLOUD-1616] Implement failure domain mapping logic

Fedosin:
When working out which machines need to be replaced/updated, the CPMS must map the machines to a failure domain index and keeping the failure domain stablethrough that action.

This commit implements this mapping functionality.


pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -82,23 +113,150 @@ func createBaseFailureDomainMapping(cpms *machinev1.ControlPlaneMachineSet, fail
...
+	if err != nil {
+		return nil, fmt.Errorf("could not convert label selector to selector: %w", err)
+	}
+
+	machineList := &machinev1beta1.MachineList{}
+	if err := cl.List(ctx, machineList, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("failed to list machines: %w", err)
+	}
+
+	failureDomains, err := providerconfig.ExtractFailureDomainsFromMachines(machineList.Items)
```
JoelSpeed, line 130: Why do you do this here rather than taking it out of the Machine when we are in the Machine loop, seems odd to get it from here instead, are we guaranteed the ordering is the same?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -82,23 +113,150 @@ func createBaseFailureDomainMapping(cpms *machinev1.ControlPlaneMachineSet, fail
...
 }
 
 // reconcileMappings takes a base mapping and a machines mapping and reconciles the differences. If any machine failure
 // domain has an identical failure domain in the base mapping, the mapping from the Machine should take precedence.
 // When overwriting a mapping, the mapping in place must be swapped to avoid losing information.
-func reconcileMappings(logger logr.Logger, base, machines map[int32]failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
+func reconcileMappings(logger logr.Logger, replicas int32, base, machines map[int32]failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
 	out := make(map[int32]failuredomain.FailureDomain)
 
+	for i := int32(0); i < replicas; i++ {
```
JoelSpeed, line 177: Why not `len(machines)`?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -82,23 +113,150 @@ func createBaseFailureDomainMapping(cpms *machinev1.ControlPlaneMachineSet, fail
...
+	sort.Slice(candidates, func(i, j int) bool { return candidates[i].String() < candidates[j].String() })
+
+	for _, candidate := range candidates {
+		if !contains(used, candidate) {
+			return &candidate
+		}
+	}
+
+	return nil
+}
```
JoelSpeed, line 262: Would like to see some additional unit tests written up for this function please

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -44,23 +50,36 @@ var (
 // then use existing Machine information to map failure domains, if possible, so that the Machine names match the
 // index of the failure domain in which they currently reside.
 func mapMachineIndexesToFailureDomains(ctx context.Context, logger logr.Logger, cl client.Client, cpms *machinev1.ControlPlaneMachineSet, failureDomains []failuredomain.FailureDomain) (map[int32]failuredomain.FailureDomain, error) {
+	out := make(map[int32]failuredomain.FailureDomain)
+
 	if len(failureDomains) == 0 {
-		return nil, errNoFailureDomains
+		logger.V(4).Info("No failure domains provided")
+
+		return out, errNoFailureDomains
```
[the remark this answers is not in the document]
Fedosin: I have to update the tests in this case as they expect to get an empty map instead of nil.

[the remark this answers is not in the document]
JoelSpeed: Ok, that's fine, let's make sure it's a nil, it's more idiomatic

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
Fedosin: Well, sometimes len(machines) !=replicas. For instance, here is an example when this happens: https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go#L810-L817

JoelSpeed: But in this logic, why do we only care about replicas, and not the complete set of Machines? I would have expected us to do the mapping over all machines. Otherwise, if there are additional machines, won't we ignore the later indexes?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -103,7 +103,7 @@ var _ = Describe("Failure Domain Mapping", func() {
 			Filters: &[]machinev1.AWSResourceFilter{
 				{
 					Name:   "tag:Name",
-					Values: []string{"subnet-us-east-1c"},
+					Values: []string{"aws-subnet-12345678"},
```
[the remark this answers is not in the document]
Fedosin: Is this important? To fix it I have to update a lot of tests by setting the correct subnet filter in all expected results.

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
Fedosin: Yes, ordering is the same. I updated this part to extract failure domain from the machine directly.

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -70,8 +89,20 @@ func mapMachineIndexesToFailureDomains(ctx context.Context, logger logr.Logger,
 func createBaseFailureDomainMapping(cpms *machinev1.ControlPlaneMachineSet, failureDomains []failuredomain.FailureDomain) (map[int32]failuredomain.FailureDomain, error) {
 	out := make(map[int32]failuredomain.FailureDomain)
 
-	// TODO: Check replicas is set, then sort the failure domains alphabetically, and use modulo arithmetic to set up the
-	// output map.
+	if cpms.Spec.Replicas == nil || *cpms.Spec.Replicas < 1 {
```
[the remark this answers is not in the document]
Fedosin: According to the test you wrote, it is considered as invalid https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go#L590-L598

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -45,22 +51,39 @@ var (
...
 
+	selector, err := metav1.LabelSelectorAsSelector(&cpms.Spec.Selector)
+	if err != nil {
+		return nil, fmt.Errorf("could not convert label selector to selector: %w", err)
+	}
+
+	machineList := &machinev1beta1.MachineList{}
+	if err := cl.List(ctx, machineList, &client.ListOptions{LabelSelector: selector}); err != nil {
+		return nil, fmt.Errorf("failed to list machines: %w", err)
+	}
```
JoelSpeed, line 67: I had expected this to be inside `createMachineMapping`, I can see you've put it here so you can get the number of machines, but I'm not sure we need that, see my other comment

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
JoelSpeed: This question is still outstanding, based on the description, I think this should be iterating over machines and checking if the index in base matches, if it doesn't, overwrite the base and swap the element. I need to take a deeper look at this logic

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
Fedosin: Done!

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -70,35 +93,210 @@ func mapMachineIndexesToFailureDomains(ctx context.Context, logger logr.Logger,
...
 // reconcileMappings takes a base mapping and a machines mapping and reconciles the differences. If any machine failure
 // domain has an identical failure domain in the base mapping, the mapping from the Machine should take precedence.
 // When overwriting a mapping, the mapping in place must be swapped to avoid losing information.
-func reconcileMappings(logger logr.Logger, base, machines map[int32]failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
+func reconcileMappings(logger logr.Logger, machineNumber int32, base, machines map[int32]failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
 	out := make(map[int32]failuredomain.FailureDomain)
+	// Consider all failure domains from machines as used.
+	used := mapToSlice(machines)
+
+	for i := int32(0); i < machineNumber; i++ {
```
JoelSpeed, line 178: This isn't necessarily true. If you have multiple machines in an index, then the number of machines could be larger than the desired output mapping.

If the desired replicas is 3, then we should have indexes 0, 1 and 2. With this logic, with 4 machines, you'd also create an index 3 which shouldn't exist.

The output mapping should be based on the `base` as that is created from the number of replicas desired, not the current desired number of machines

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -562,6 +650,180 @@ var _ = Describe("Failure Domain Mapping", func() {
 					},
 				},
 			}),
+			Entry("with machines in not matched by the selector, they are ignored", mappingMachineIndexesTableInput{
```
JoelSpeed, line 653: This test isn't correct, you're saying that the entire mapping should only return one index mapping, yet, the CPMS defines 3 failure domains and has 3 replicas. In this case, I would expect 3 indexes to be in the output mapping

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -562,6 +650,180 @@ var _ = Describe("Failure Domain Mapping", func() {
...
+					machineBuilder.WithName("machine-a").WithProviderSpecBuilder(usEast1aProviderSpecBuilder).Build(),
+					machineBuilder.WithName("machine-1").WithProviderSpecBuilder(usEast1bProviderSpecBuilder).Build(),
+					machineBuilder.WithName("machine-2").WithProviderSpecBuilder(usEast1cProviderSpecBuilder).Build(),
+					machineBuilder.WithName("machine-3").WithProviderSpecBuilder(usEast1aProviderSpecBuilder).Build(),
+				},
+				expectedMapping: map[int32]failuredomain.FailureDomain{
+					0: failuredomain.NewAWSFailureDomain(usEast1aFailureDomainBuilder.Build()),
+					1: failuredomain.NewAWSFailureDomain(usEast1bFailureDomainBuilder.Build()),
+					2: failuredomain.NewAWSFailureDomain(usEast1cFailureDomainBuilder.Build()),
+					3: failuredomain.NewAWSFailureDomain(usEast1aFailureDomainBuilder.Build()),
```
JoelSpeed, line 697: There would only be a third index if the CPMS had 4 replicas

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -562,6 +650,180 @@ var _ = Describe("Failure Domain Mapping", func() {
...
+					machineBuilder.WithName("machine-a").WithProviderSpecBuilder(usEast1aProviderSpecBuilder).Build(),
+					machineBuilder.WithName("machine-b").WithProviderSpecBuilder(usEast1bProviderSpecBuilder).Build(),
+					machineBuilder.WithName("machine-c").WithProviderSpecBuilder(usEast1cProviderSpecBuilder).Build(),
+					machineBuilder.WithName("machine-0").WithProviderSpecBuilder(usEast1aProviderSpecBuilder).Build(),
+				},
+				expectedMapping: map[int32]failuredomain.FailureDomain{
+					0: failuredomain.NewAWSFailureDomain(usEast1aFailureDomainBuilder.Build()),
+					1: failuredomain.NewAWSFailureDomain(usEast1bFailureDomainBuilder.Build()),
+					2: failuredomain.NewAWSFailureDomain(usEast1cFailureDomainBuilder.Build()),
+					3: failuredomain.NewAWSFailureDomain(usEast1aFailureDomainBuilder.Build()),
```
JoelSpeed, line 738: There would only be a third index if the cpms had 4 replicas

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -99,6 +175,196 @@ func createMachineMapping(ctx context.Context, logger logr.Logger, cl client.Cli
...
+	// As a key we use failure domain string representation.
+	failureDomainCount := make(map[string]int)
+
+	for _, fd := range failureDomains {
+		stringsToFailureDomains[fd.String()] = fd
+		failureDomainCount[fd.String()] = 0
+	}
+
+	// maxCount shows the allowed amount for each failure domain in the mapping.
+	maxCount := len(machines)/len(failureDomains) + 1
```
JoelSpeed, line 328: What would happen here if the number of machines or failure domains isn't odd, eg 3 machines with 1 failure domain, this becomes 3/2 which becomes 1, and then I'm not sure the balancing would work, you actually want 3/1 there

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -99,6 +175,196 @@ func createMachineMapping(ctx context.Context, logger logr.Logger, cl client.Cli
...
+	// stringsToFailureDomains is used to quickly convert failure domain string representation into the failure domain.
+	stringsToFailureDomains := make(map[string]failuredomain.FailureDomain)
+
+	// failureDomainCount contains the amount for each failure domain in machines.
+	// As a key we use failure domain string representation.
+	failureDomainCount := make(map[string]int)
+
+	for _, fd := range failureDomains {
+		stringsToFailureDomains[fd.String()] = fd
+		failureDomainCount[fd.String()] = 0
```
JoelSpeed, line 324: Not sure how I feel about the mapping based on the string representation here, would it be better to use some helper functions to match the failuredomains based on their own Equal function for comparison vs the string comparison?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -82,14 +110,62 @@ func createBaseFailureDomainMapping(cpms *machinev1.ControlPlaneMachineSet, fail
...
+			// Ignore the machine as it doesn't contain an index in its name.
+			logger.V(4).Info(
+				"Ignoring machine in failure domain mapping with unexpected name",
+				"machine", machine.Name,
+			)
+
+			continue
+		}
+
+		if fd, ok := out[int32(machineNameIndex)]; ok && fd.String() != failureDomain.String() {
```
JoelSpeed, line 144: We have a function for comparing these
```suggestion
		if fd, ok := out[int32(machineNameIndex)]; ok && !fs.Equal(failureDomain) {
```

JoelSpeed:
This is continued in #69

PR #69, opened 2022-07-26
[OCPCLOUD-1616] Implement failure domain mapping logic

JoelSpeed:
This is a continuation of #62.

While reviewing and working with @Fedosin on the implementation, we added a bunch of new tests and found various edge cases and evolved the implementation.
This is our most complex logic within the operator (in particular the reconcileMappings function) so I want to make sure that what we have is well tested and readable/understandable for new consumers.

After much deliberation and consultation with @damdo and @elmiko I think we should go with this approach for the implementation


damdo (commented):
/lgtm

PR #70, opened 2022-07-26
Remove redundant "template" in error path

RadekManak:
During QE review, it was noticed that an extraneous `template` was added into the path of this error


damdo (commented):
/lgtm

PR #71, opened 2022-07-29
Add missing "spec" in error path

RadekManak:
"template" field in error path was erroneously removed instead of changing it into "spec".
This PR adds the "spec" field to the error path.


JoelSpeed (commented):
/approve
/lgtm

PR #73, opened 2022-08-03
Ensure additional failure domains do not prevent installation of a ControlPlaneMachineSet

JoelSpeed:
The failure domain checks within the webhooks are an attempt to prevent immediate rollouts when a ControlPlaneMachineSet is installed. Previously, we have said that the list of failure domains must match exactly with the list of failure domains present within Control Plane Machines.

However, in some regions, the installer would normally configure additional failure domains (eg us-east-1a which has more than 3 failure domains). In this case, it is not harmful to contain the additional failure domains as this would allow balancing later if the control plane were horizontally scaled.

Therefore, this PR changes the behaviour to only prevent the request if there are failure domains that are unused and failure domains that are duplicated within the Machines. This should mean additional failure domains can be added, but will prevent immediate rollouts when the ControlPlaneMachineSet would otherwise rebalance the Control Plane Machines.


lobziik:
linters failed, otherwise lgtm

PR #75, opened 2022-08-04
Test CPMS degraded when failing to create new Machines

JoelSpeed:
If we know that we aren't going to progress with a rollout, we should degrade the operator.
So, if we can see that a newly created replacement machine errors straight away, for example because the new instance type was invalid, then we should report that and go degraded.

If an existing Machine goes failed, this wouldn't result in degraded as a MachineHealthCheck should cause the Machine to be deleted, and then in that case, we would replace the Machine. Also, if we do go degraded when any Machine is failed, we will not be able to create a replacement Machine if the Failed Machine becomes deleted.


damdo (commented):
Looks a reasonable behaviour.
Thanks for adding this and the test.
/lgtm

PR #76, opened 2022-08-04
Add OpenShift standard leader election configuration to manager

JoelSpeed:
As a core OpenShift operator, we must respect the leader election conventions and run the manager with leader election.
This PR sets up the leader election config based on what we do in the CCCMO which was the most recently implemented.


cmd/control-plane-machine-set-operator/main.go
```diff
@@ -52,30 +57,48 @@ func main() { //nolint:funlen
...
-		metricsAddr          string
-		enableLeaderElection bool
-		probeAddr            string
+		metricsAddr      string
+		probeAddr        string
+		managedNamespace string
+
+		leaderElectionConfig = config.LeaderElectionConfiguration{
+			LeaderElect:  true,
+			ResourceName: "control-plane-machine-set-leader",
```
lobziik, line 66: nit: suggest to move this resource name to constant

damdo (approved):
/lgtm

PR #77, opened 2022-08-04
Add image-references

Fedosin:
To allow CPMS operator to be installed by CVO we need to create an
image-references file.

For more information:
https://github.com/openshift/enhancements/blob/master/dev-guide/cluster-version-operator/dev/operators.md#how-do-i-ensure-the-right-images-get-used-by-my-manifests

### DIFF
--- a/manifests/0000_31_control-plane-machine-set-operator_03_deployment.yaml
+++ b/manifests/0000_31_control-plane-machine-set-operator_03_deployment.yaml
@@ -24,7 +24,7 @@ spec:
       serviceAccountName: control-plane-machine-set-operator
       containers:
       - name: control-plane-machine-set-operator
-        image: quay.io/origin/origin-control-plane-machine-set-operator
+        image: quay.io/openshift/origin-cluster-control-plane-machine-set-operator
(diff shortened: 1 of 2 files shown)


JoelSpeed:
/hold

It's my understanding that we need to have the images in payload before we add an image reference, so lets make sure we are doing that and then we can proceed to work out why the CPMSO isn't being deployed

Fedosin:
@JoelSpeed, no, it's not related... For instance, for CCCMO we created the file on Jan 19, 2021: https://github.com/openshift/cluster-cloud-controller-manager-operator/pull/3 but the image was added 6 months later, on Jun 18, 2021: https://github.com/openshift/ocp-build-data/pull/936

JoelSpeed:
I still don't see the operator being deployed in the CI builds though, any ideas?

Fedosin:
We need to merge it first. On the next CI run we will see the image deployed.

JoelSpeed:
Changes in a PR should have an effect on the CI before they are merged, I'm not sure what you're saying is true here

PR #78, opened 2022-08-04
[OCPCLOUD-1650] Ensure we replace deleted machines in the RollingUpdate strategy

JoelSpeed:
We need to make sure that if there is a machine that has been marked for deletion, that we replace it appropriately.

Therefore, when creating machines, we must include deleting machines from the those we think need replacement and we must exclude deleting machines from the list we think could be replacing an already deleting machine.

This PR updates the rolling update strategy and tests to account for this.

### DIFF
--- a/pkg/controllers/controlplanemachineset/updates.go
+++ b/pkg/controllers/controlplanemachineset/updates.go
@@ -204,7 +204,7 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineOnDeleteUpdate(ctx co
 
 func (r *ControlPlaneMachineSetReconciler) waitForPendingMachines(logger logr.Logger, machines []machineproviders.MachineInfo) bool {
 	machinesPending := pendingMachines(machines)
-	machinesNeedingUpdate := needUpdateMachines(machines)
+	machinesNeedingReplacement := needReplacementMachines(machines)
 	machinesReady := readyMachines(machines)
 
 	// Find out if and what Machines in this index need an update.
@@ -219,13 +219,13 @@ func (r *ControlPlaneMachineSetReconciler) waitForPendingMachines(logger logr.Lo
 		return true
 	}
 
-	if hasAny(machinesNeedingUpdate) && hasAny(machinesPending) {
+	if hasAny(machinesNeedingReplacement) && hasAny(machinesPending) {
 		// A Pending Machine Replacement already exists.
 		// Wait for it to become Ready.
 		// Consider the first found pending machine for this index to be the replacement machine.
 		replacementMachine := machinesPending[0]
 		// Consider the first found outdated machine for this index to be the one in need of update.
-		outdatedMachine := machinesNeedingUpdate[0]
+		outdatedMachine := machinesNeedingReplacement[0]
 
 		logger := logger.WithValues("index", int(outdatedMachine.Index), "namespace", r.Namespace, "name", outdatedMachine.MachineRef.ObjectMeta.Name)
 		logger.V(2).WithValues("replacementName", replacementMachine.MachineRef.ObjectMeta.Name).Info(waitingForReplacement)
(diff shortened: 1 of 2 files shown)


pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -402,6 +404,20 @@ func updatedMachines(machinesInfo []machineproviders.MachineInfo) []machineprovi
 	return result
 }
 
+// updatedMachines returns the list of MachineInfo which have an Updated (Spec up-to-date and Ready) Machine and
```
damdo, line 407: nit, this should match the func name.

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -402,6 +404,20 @@ func updatedMachines(machinesInfo []machineproviders.MachineInfo) []machineprovi
...
+	result := []machineproviders.MachineInfo{}
+
+	for i := range machinesInfo {
+		if machinesInfo[i].Ready && !machinesInfo[i].NeedsUpdate && !isDeletedMachine(machinesInfo[i]) {
+			result = append(result, machinesInfo[i])
+		}
+	}
+
+	return result
+}
```
damdo, line 419: In spirit with the chaining nature of these functions I'd
call it `nonDeletedMachines` just checking for `!isDeletedMachine(machinesInfo[i])`
Then later when using them, we can do:
```
nonDeletedMachines(updatedMachines(machines))
```
So then both can be used independently or in combination :)

I'd choose `Deleted` as opposed to `Terminated` for this use case, because usually Deleted is referring to objects with `DeletionTimestamp != nil` while `Terminated` is mostly used for when backing cloud instances are actually gone.

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -283,9 +283,9 @@ func (r *ControlPlaneMachineSetReconciler) deleteReplacedMachines(ctx context.Co
 }
 
 func (r *ControlPlaneMachineSetReconciler) createReplacementMachines(ctx context.Context, logger logr.Logger, machineProvider machineproviders.MachineProvider, machines []machineproviders.MachineInfo, idx int, maxSurge int, surgeCount *int) (bool, ctrl.Result, error) {
-	machinesNeedingUpdate := needUpdateMachines(machines)
+	machinesNeedingUpdate := needReplacementMachines(machines)
```
damdo, line 286: I'd then call `machinesNeedingUpdate` => `machinesNeedingReplacement` everywhere.

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -863,6 +863,111 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+								"updateStrategy", machinev1.RollingUpdate,
+								"index", 1,
+								"namespace", namespaceName,
+								"name", "machine-1",
+							},
+							Message: waitingForRemoved,
+						},
+					}
+				},
+			}),
```
damdo, line 970: I think for this one we already have a test here: https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/17d16b6c891ef8afd0c6c810c58a5129a56806f1/pkg/controllers/controlplanemachineset/updates_test.go#L300-L327

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -863,6 +863,111 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+						},
+						{
+							Level: 2,
+							KeysAndValues: []interface{}{
+								"updateStrategy", machinev1.RollingUpdate,
+								"index", 1,
+								"namespace", namespaceName,
+								"name", "machine-1",
+							},
+							Message: createdReplacement,
```
damdo, line 897: As discussed on the call, it would be nice to have a comment (on this and the next test) explaining why on an index with `waitingForRemoved`  where we are not supposed to proceed further, we instead do (i.e. mentioning the etcd deletion hooks matter).

damdo (commented):
Thank you for making the mechanism robust to this corner case.
Left a couple of minor things but overall it LGTM

pkg/controllers/controlplanemachineset/updates_test.go
JoelSpeed: This test has a different of the machines needing update no? Like it has an index with an update, where this new test doesn't, so slightly different?

damdo: Ah yes right! Ok never mind then.

pkg/controllers/controlplanemachineset/updates.go
JoelSpeed: I think that would mean running through the loop again right? If we want to make this more composable I think we can create different composable filters and make it so that we only run the loop once.

For now I think lets leave this as is and discuss a refactor later

JoelSpeed:
Thanks for the review @damdo, let me know if you think the comments I've added capture the sentiment we discussed on the call

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -283,9 +283,9 @@ func (r *ControlPlaneMachineSetReconciler) deleteReplacedMachines(ctx context.Co
 }
 
 func (r *ControlPlaneMachineSetReconciler) createReplacementMachines(ctx context.Context, logger logr.Logger, machineProvider machineproviders.MachineProvider, machines []machineproviders.MachineInfo, idx int, maxSurge int, surgeCount *int) (bool, ctrl.Result, error) {
-	machinesNeedingUpdate := needUpdateMachines(machines)
+	machinesNeedingReplacement := needReplacementMachines(machines)
 	machinesPending := pendingMachines(machines)
-	machinesUpdated := updatedMachines(machines)
+	machinesUpdatedNonTerminated := updatedNonDeletedMachines(machines)
```
damdo, line 288: Should we call the var on the lefthand side `machinesUpdatedNonDeleted` here then?

damdo (commented):
Added a minor nit, otherwise
/lgtm

damdo (commented):
/lgtm

PR #79, opened 2022-08-05
Bug 2115638: Ensure that the failure domain mapping is respected in getMachineInfo

JoelSpeed:
This fixes two things, firstly that we weren't respecting the failure domain mapping from reconcile mappings when a Machine had a different failure domain. This means the rebalancing won't work. We were comparing the failure domain from the machine to failure domain from the machine, where it should have been the failure domain from the mapping.

Secondly, for some reason logging the mapping directly resulted in a map of integers to nil values, the only way I could work out to fix this was to flatten it to a string.

This PR would best be reviewed commit wise

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -57,6 +57,12 @@ var (
 	// not currently specified in the ControlPlaneMachineSet definition. User intervention is required here.
 	errCouldNotDetermineMachineIndex = errors.New("could not determine Machine index from name or failure domain")
 
+	// errCouldNotFindFailureDomain is used to denote that the MachineProvider could not find a failure domain
+	// within the mapping for a given index.
+	// This means the Machine has an index in its name that is outside the bounds of the current mapping.
+	// This may happen if the cluster is trying to horizontally scale down.
+	errCouldNotFindFailureDomain = errors.New("could not find failure domain for index")
+
 	// errEmptyConfig is used to denote that the machine provider could not be constructed
 	// because no configuration was provided by the user.
 	errEmptyConfig = fmt.Errorf("cannot initialise %s provider with empty config", machinev1.OpenShiftMachineV1Beta1MachineType)
@@ -180,7 +186,7 @@ func (m *openshiftMachineProvider) GetMachineInfos(ctx context.Context, logger l
 	}
 
 	for _, machine := range machineList.Items {
-		machineInfo, err := m.generateMachineInfo(machine, logger)
+		machineInfo, err := m.generateMachineInfo(logger, machine)
 		if err != nil {
 			return nil, fmt.Errorf("could not generate machine info for machine %s: %w", machine.Name, err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -267,12 +250,45 @@ func (m *openshiftMachineProvider) generateMachineInfo(machine machinev1beta1.Ma
...
+		NeedsUpdate:  !configsEqual,
+		Index:        machineIndex,
 		ErrorMessage: pointer.StringDeref(machine.Status.ErrorMessage, ""),
 	}, nil
 }
 
+func (m *openshiftMachineProvider) getMachineIndex(logger logr.Logger, machine machinev1beta1.Machine) (int32, error) {
+	machineNameIndex, correctFormat := getMachineNameIndex(machine)
+	if correctFormat {
+		// If the machine name has the correct format we implicitly trust it to be correct.
```
lobziik, line 262: I thinking about add an extra check there for such index existence.
WDYT about move the check from line 233 inside this function?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -267,12 +250,45 @@ func (m *openshiftMachineProvider) generateMachineInfo(machine machinev1beta1.Ma
...
 		Ready:        pointer.StringDeref(machine.Status.Phase, "") == runningPhase,
-		NeedsUpdate:  !configsEqual || !indexFound,
-		Index:        index,
+		NeedsUpdate:  !configsEqual,
+		Index:        machineIndex,
 		ErrorMessage: pointer.StringDeref(machine.Status.ErrorMessage, ""),
 	}, nil
 }
 
+func (m *openshiftMachineProvider) getMachineIndex(logger logr.Logger, machine machinev1beta1.Machine) (int32, error) {
```
lobziik, line 259: WDYT about name this function as "getMachineFailureDomainIndex"? Or this index might be used for smth else?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: If we move the check, we still need to have a check when we extract from the map anyway. I would want to blindly take an object from the map and not check that the key was present. So I think for now, keep this function as getting the index, and the outer function checking that it's correct

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: The index is related to the failure domain yeah but it's also how we refer to each of the machines and how we name them, so I think we should leave as is on this one. For example if it were `getMachineFailureDomainIndex` I would expect the function to be looking up the index from the failure domain on the machine, but that's the fallback logic here

damdo (commented):
/lgtm

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
lobziik: Dont like that fallback inside, tbh. Looks a bit obscured to me. But ok, lets leave it as is

PR #80, opened 2022-08-08
Add manifests in the resulting docker image

Fedosin:
Currently we don't copy manifests in the image and therefore CVO can't deploy it.

### DIFF
--- a/cmd/control-plane-machine-set-operator/main.go
+++ b/cmd/control-plane-machine-set-operator/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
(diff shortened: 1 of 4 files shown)


JoelSpeed:
This is failing because we don't have the version set in the cluster operator. Let's make sure we fix that before merging this
/hold

cmd/control-plane-machine-set-operator/main.go
```diff
@@ -166,3 +172,13 @@ func setupScheme(scheme *runtime.Scheme) error {
 
 	return nil
 }
+
+func getReleaseVersion() string {
+	releaseVersion := os.Getenv(releaseVersionEnvVariableName)
+	if len(releaseVersion) == 0 {
+		releaseVersion = unknownVersionValue
+		klog.Infof("%s environment variable is missing, defaulting to %q", releaseVersionEnvVariableName, unknownVersionValue)
```
JoelSpeed, line 180: Please pass in a `logr.Logger` and log using that to be consistent with the rest of this operator. You can pass through the `setupLog` we init on line 57

PR #82, opened 2022-08-10
Bug 2115358: Fix panic when injecting nil FailureDomain

RadekManak:
This PR fixes panic when trying to inject nil failure domain and two potential panics when comparing FailureDomains and ProviderConfigs.


JoelSpeed (commented):
/lgtm
/approve

PR #83, opened 2022-08-15
Bug 2115358: Ensure machine provider info can be gathered in a single failure domain

JoelSpeed:
When gathering Machine Provider information, we should ignore the failure domain comparison when no failure domains were provided in the CPMS. The Machine provider will have no failure domains set in this case so we can use a check against the length of the failure domains to verify this.

This is required to support single availability zone depoyments

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -230,15 +230,21 @@ func (m *openshiftMachineProvider) generateMachineInfo(logger logr.Logger, machi
 		return machineproviders.MachineInfo{}, fmt.Errorf("could not compare existing and desired provider configs: %w", err)
 	}
 
-	// Make sure to compare using the desired failure domain from the mapping.
-	mappedFailureDomain, ok := m.indexToFailureDomain[machineIndex]
-	if !ok {
-		return machineproviders.MachineInfo{}, fmt.Errorf("%w: unknown index %d", errCouldNotFindFailureDomain, machineIndex)
-	}
+	templateProviderConfig := m.providerConfig
 
-	templateProviderConfig, err := m.providerConfig.InjectFailureDomain(mappedFailureDomain)
-	if err != nil {
-		return machineproviders.MachineInfo{}, fmt.Errorf("error injecting failure domain into provider config: %w", err)
+	if len(m.indexToFailureDomain) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -230,15 +230,21 @@ func (m *openshiftMachineProvider) generateMachineInfo(logger logr.Logger, machi
...
-	mappedFailureDomain, ok := m.indexToFailureDomain[machineIndex]
-	if !ok {
-		return machineproviders.MachineInfo{}, fmt.Errorf("%w: unknown index %d", errCouldNotFindFailureDomain, machineIndex)
-	}
+	templateProviderConfig := m.providerConfig
 
-	templateProviderConfig, err := m.providerConfig.InjectFailureDomain(mappedFailureDomain)
-	if err != nil {
-		return machineproviders.MachineInfo{}, fmt.Errorf("error injecting failure domain into provider config: %w", err)
+	if len(m.indexToFailureDomain) > 0 {
```
lobziik, line 235: I'm thinking about adding a log messages there which explains what is going on with failure domains.
Like "No failure domains found" and "Failure domains found, updating provider config"

lobziik:
aside my comment about logging - looks good to me

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: We already have logging at level 4 after this point that logs which machines are mapped to which indexes and which indexes are mapped to which failure domains so I think we would probably just be duplicating information in that case

https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/44c90332eefb8504d78c32bdf77dab86c81a583a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go#L198-L213

https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/44c90332eefb8504d78c32bdf77dab86c81a583a/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go#L72-L75

lobziik: Cool. Second one which you pointed should be suffice i think. Thanks

JoelSpeed:

Failures appear to be unrelated at there's no CPMS resource present in the test cluster right now

PR #84, opened 2022-08-17
Bug 2117439: Add webhook validation to prevent scenarios which can create known broken clusters

JoelSpeed:
We have identified via this bug, that there are two scenarios in which an automated replacement of Control Plane Machines can lead to a broken cluster. Both related to load balancers.

On Azure, the `internalLoadBalancer` has historically not been set by the installer. This is needed (and supported) by Machine API to ensure that the load balancer used by control plane components and kubelet has backends to fulfil the requests. Without this parameter, the CPMS will happily replace all Machines and the cluster will grind to a halt as KCM will lose communication within the API.

On GCP, we have a similar scenario, except updating the internal load balancer backend pool is not currently supported by Machine API on GCP. Until that issue is resolved, we must prevent users installing a CPMS on GCP.

### DIFF
--- a/pkg/webhooks/controlplanemachineset/webhooks.go
+++ b/pkg/webhooks/controlplanemachineset/webhooks.go
@@ -246,6 +246,7 @@ func validateOpenShiftMachineV1BetaTemplate(parentPath *field.Path, template mac
 
 	errs = append(errs, validateTemplateLabels(parentPath.Child("metadata", "labels"), template.ObjectMeta.Labels, selector)...)
 	errs = append(errs, validateOpenShiftFailureDomains(parentPath.Child("failureDomains"), template.FailureDomains)...)
+	errs = append(errs, validateOpenShiftProviderConfig(parentPath, template)...)
 
 	return errs
 }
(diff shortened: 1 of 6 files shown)


pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -337,6 +338,48 @@ func validateOpenShiftAWSFailureDomains(parentPath *field.Path, failureDomains *
...
+	config := providerConfig.Config()
+
+	if config.InternalLoadBalancer == "" {
+		errs = append(errs, field.Required(parentPath.Child("internalLoadBalancer"), "internalLoadBalancer is required for control plane machines"))
+	}
+
+	return errs
+}
+
+// validateOpenShiftAzureProviderConfig runs GCP specific checks on the provider config on the ControlPlaneMachineSet.
```
damdo, line 375:
```suggestion
// validateOpenShiftGCPProviderConfig runs GCP specific checks on the provider config on the ControlPlaneMachineSet.
```

JoelSpeed: Good spot, resolved

damdo (commented):
/lgtm

JoelSpeed:
/hold cancel
/retest-required

Thanks @huali9

PR #86, opened 2022-08-18
[OCPCLOUD-1647] Implement pre-flight Cluster state validation

damdo:
We need to make sure a number of checks pass before the CPMS controller can perform any operations.
The things to be checked are:

1. **All Nodes in the cluster claiming to be control plane nodes have a valid machine**.
1. **At least 1 of the control plane machines is in the ready state** (if there are no ready Machines then the cluster is likely misconfigured).
1. **The cluster has the correct number of indexes**:
      + Right number of indexes, The cluster state is valid.
      + Too few indexes, The cluster state is valid. We will later scale up without user intervention when we perform reconcileMachineUpdates.
      + Too many indexes, invalid. We set the operator to degraded and ask the user for manual intervention.
1. **No replacement machines** (one that doesn't need update but has an equivalent in the index that needs update) **have an error.**

If any of these conditions is not met, the cpms operator will go degraded and will halt all operations until the issue has been resolved.

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
@@ -27,9 +28,11 @@ import (
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	errorutils "k8s.io/apimachinery/pkg/util/errors"
@@ -56,6 +59,20 @@ const (
 	// degradedClusterState is used to denote that the control plane machine set has detected a degraded cluster.
 	// In this case, the controller will not perform any further actions.
 	degradedClusterState = "Cluster state is degraded. The control plane machine set will not take any action until issues have been resolved."
+)
+
+var (
+	// errFoundUnmanagedControlPlaneNodes is used to inform users that one or more nodes do not have machines associated with them.
+	errFoundUnmanagedControlPlaneNodes = errors.New("found unmanaged control plane nodes, the following node(s) do not have associated machines")
+
+	// errNoReadyControlPlaneMachines is used to inform users that no control plane machines in the cluster are ready.
+	errNoReadyControlPlaneMachines = errors.New("no ready control plane machines")
+
+	// errFoundErroredReplacementControlPlaneMachine is used to inform users that one or more replacement control plane machines, have been found.
+	errFoundErroredReplacementControlPlaneMachine = errors.New("found replacement control plane machines in an error state, the following machines(s) are currently reporting an error")
+
+	// errFoundExcessiveIndexes is used to inform users that an excessive number of indexes has been found.
+	errFoundExcessiveIndexes = errors.New("found an excessive number of indexes for the control plane machine set")
 )
 
 // ControlPlaneMachineSetReconciler reconciles a ControlPlaneMachineSet object.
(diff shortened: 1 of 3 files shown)


pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,181 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
...
+		nodesList := &corev1.NodeList{}
+		if err := r.List(ctx, nodesList, &client.ListOptions{
+			LabelSelector: labels.SelectorFromSet(map[string]string{
+				label: "",
+			}),
+		}); err != nil {
+			return false, fmt.Errorf("failed to get Nodes: %w", err)
+		}
+
+		cpmsNodes = append(cpmsNodes, nodesList.Items...)
```
JoelSpeed, line 550: Does this mean we could end up with duplicate nodes? Should we deduplicate this list based on node names?

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,181 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
 
 	return false
 }
+
+func (r *ControlPlaneMachineSetReconciler) checkControlPlaneNodesToMachinesMappings(ctx context.Context, logger logr.Logger,
```
JoelSpeed, line 533: Missing description

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,181 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
...
+			Message: fmt.Sprintf("Found %d unmanaged node(s)", len(unmanagedNodeNames)),
+		})
+
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (r *ControlPlaneMachineSetReconciler) checkReadyControlPlaneMachineExists(logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, sortedIndexedMs [][]machineproviders.MachineInfo) bool {
```
JoelSpeed, line 596: Missing description

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,181 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
...
+			Message: "No ready control plane machines found",
+		})
+
+		return false
+	}
+
+	return true
+}
+
+func (r *ControlPlaneMachineSetReconciler) checkCorrectNumberOfIndexes(logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, sortedIndexedMs [][]machineproviders.MachineInfo) bool {
```
JoelSpeed, line 631: Missing description

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,181 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
...
+			Message: fmt.Sprintf("Observed %d index(es) in excess", excessiveIndexes),
+		})
+
+		return false
+	}
+
+	return true
+}
+
+func (r *ControlPlaneMachineSetReconciler) checkNoErrorForReplacements(logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, sortedIndexedMs [][]machineproviders.MachineInfo) bool {
```
JoelSpeed, line 669: Missing description

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,181 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
...
+	var nonReadyReplacementMachineNames []string
+
+	for _, machines := range sortedIndexedMs {
+		machinesUpdated := updatedMachines(machines)
+		machinesOutdated := needReplacementMachines(machines)
+
+		if hasAny(machinesOutdated) && hasAny(machinesUpdated) {
+			for _, m := range machinesUpdated {
+				if m.ErrorMessage != "" {
+					nonReadyReplacementMachineNames = append(nonReadyReplacementMachineNames, m.MachineRef.ObjectMeta.Name)
```
JoelSpeed, line 679: Are these nonReady or errored? I think nonReady would be valid no?

pkg/controllers/controlplanemachineset/controller.go
damdo: Yes better. Done.

pkg/controllers/controlplanemachineset/controller.go
damdo: Ah good catch! Updated.

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -470,3 +529,196 @@ func isOwnedByCurrentCPMS(cpms *machinev1.ControlPlaneMachineSet, machineObjectM
...
+		})
+
+		return false
+	}
+
+	return true
+}
+
+// fetchControlPlaneNodes fetches a map of unique nodes that have the "control-plane" (and/or legacy "master") labels.
+func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Context) (*map[string]corev1.Node, error) {
```
JoelSpeed, line 707: Why return a pointer to a map?

damdo: It is superfluous for maps, you are right. Removed

JoelSpeed:
Ok, looks good, please squash and I'll label

PR #87, opened 2022-08-24
add more options for running tests

elmiko:
This change adds a new Makefile target named `test-only` that will run
just the tests with no linting or vendoring activities. It also adds the
`GINKGO_EXTRA_ARGS` environment variable to the `hack/test.sh` script so
that users can inject extra commands to the ginkgo runner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,10 @@ lint: ## Run golangci-lint over the codebase.
 	./hack/verify-log-keys.sh
 
 .PHONY: test
-test: generate fmt vet ## Run tests.
+test: generate fmt vet unit ## Run tests.
+
+.PHONY: unit
+unit: ## Run only the tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin)" ./hack/test.sh
 
 .PHONY: verify-%
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ Tooling is vendored and executed using `go run` so no additional tooling should 
 * `make build`: Build the operator binary into `bin/manager`
 * `make test`: Run the project tests. Tests are written using ginkgo and the ginkgo tooling is used to execute the
 tests. In CI, this target outputs JUnit and code coverage reports. Depends on `generate`, `fmt` and `vet` tasks.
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -74,6 +74,10 @@ lint: ## Run golangci-lint over the codebase.
 test: generate fmt vet ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path --bin-dir $(PROJECT_DIR)/bin)" ./hack/test.sh
```
JoelSpeed, line 75: To avoid drift between this and the test-only target, we can make test rely on the test-only target
```suggestion
test: generate fmt vet test-only ## Run tests.
```

Makefile
```diff
@@ -74,6 +74,10 @@ lint: ## Run golangci-lint over the codebase.
...
 
+.PHONY: test-only
+test-only: ## Run only the tests.
```
JoelSpeed, line 78: On other repos, I think we call this make target `unit`, wdyt?

damdo: +1

Makefile
elmiko: yeah, makes sense

Makefile
elmiko: sure thing

elmiko:
updated to address comments

README.md
```diff
@@ -51,6 +51,8 @@ Tooling is vendored and executed using `go run` so no additional tooling should
 * `make build`: Build the operator binary into `bin/manager`
 * `make test`: Run the project tests. Tests are written using ginkgo and the ginkgo tooling is used to execute the
 tests. In CI, this target outputs JUnit and code coverage reports. Depends on `generate`, `fmt` and `vet` tasks.
+* `make test-only`: Run only the project tests, without the `generate`, `fmt`, or `vet` tasks. The `GINKGO_EXTRA_ARGS`
```
damdo, line 54:
```suggestion
* `make unit`: Run only the project tests, without the `generate`, `fmt`, or `vet` tasks. The `GINKGO_EXTRA_ARGS`
```

elmiko: weird, i thought i had updated this but it didn't get pushed.

elmiko:
updated the missed readme info

JoelSpeed (commented):
/approve
/lgtm

JoelSpeed:

This is only a makefile change

PR #92, opened 2022-09-05
[OCPCLOUD-1570] Add initial end user documentation

JoelSpeed:
In the interest of making the reviews easier, I'm splitting the work from this card into several PRs. This PR contains the first sections which contains an overview/summary of the CPMS, a quick look at some of the features, a start of a glossary and some discussion about how the rolling update and on delete strategies work.

I have a list of other things to include in follow up PRs which I'll push once this stuff is reviewed:
- [ ] More detail on failure domains
  - [ ] How do the failure domains relate to cloud provider concepts
  - [ ] How to configure failure domains on each provider
- [ ] How does this fit in with the rest of Machine API
  - [ ] Differences between CPMS and regular MachineSets
  - [ ] Why do we have failure domains
  - [ ] Why do we have rolling update strategies
- [ ] Troubleshooting guide
  - [ ] Examples of when the operator will go degraded, and why
  - [ ] Corner cases
- [ ] FAQs
  - [ ] Connection to Cluster API
  - [ ] How are providers integrated


docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Features
+
+### Spreading machines across Failure Domains
+
+The control plane machine set can balance machines across multiple failure domains. This provides high availability and
+fault tolerance within the control plane when issues arise within the infrastructure provider.
+
+Failure domains are configured within the machine template of the control plane machine set and are currently supported
+on Amazon Web Services (AWS) and Microsoft Azure only. Other platforms will be added over time.
```
elmiko, line 67: mentioning AWS and Azure makes me think we might want a feature matrix at some point, to avoid the need for updating the text of the docs.

not a blocker, just a suggestion

docs/user/installation.md
```diff
@@ -0,0 +1,147 @@
+# Installation
+
+This document provides instructions for users to configure a control plane machine set within an OpenShift cluster.
+This process can be followed on OpenShift clusters that are version 4.12 or higher.
+
+## Installer based installation
+
```
elmiko, line 7: is there going to be some text here in the future?

just wondering if we should put a "TBD" here, or maybe just a simple sentence or something,

docs/user/update-strategies.md
```diff
@@ -0,0 +1,103 @@
...
+Once the replacement machine is ready (synonymous with the `Running` phase in Machine API), the control plane machine
+set will delete the old machine.
+This deletion will then signal to the etcd operator to move the etcd member from the old machine to the new machine.
+
+Once the etcd operator has moved the etcd member, it will remove the machine deletion hook from the machine which will
+then allow the machine to be removed from the cluster by the machine controller.
+
+The control plane machine set, on each reconcile, iterates through the machine indexes applying this logic.
+Where the control plane machine set differs from a deployment is that the `maxSurge` concept of the deployment, which
+allows over-provisioning of the workload during an update, is limited to 1 in the control plane machine set.
```
elmiko, line 27: i might throw a code block around the "1" to indicate that it is an exact value to be input

elmiko (commented):
reads well to me, i have a couple suggestions but they are not blockers
/lgtm

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+A failure domain defines a fault boundary within the infrastructure provider. Failure domains are also known as zones
+or availability zones depending on the infrastructure provider.
+
+### Index
+
+Each control plane machine is considered to exist within an index of the control plane.
+The index of each machine is typically the last digit of the machine name.
+Control plane machines are indexed from 0, so are typically indexed as 0, 1 and 2 (and 3 and 4 in the case of a 5
+member control plane).
```
lobziik, line 131: Not actionable, just a note. 

I found a bit difficult to comprehend what is the "index" here. Don't know yet how to describe it better, maybe worth to outline that it's the internal (exists only in the context of this controller) approach to group/mark machines.

Maybe some recap of the problem area would help here, like a notion that we need to replace exact machine within the same failure domain.

lobziik (commented):
couple of notes, but not a blockers

/lgtm

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+The control plane machine set constantly monitors the control plane machines within the cluster and compares their
+configured specification with that of the machine template embedded within it.
+When a machine is considered to need an update (either the specification differs or it has been deleted), the control
+plane machine set will replace the machine with an update instance based on the update strategy defined within the
+control plane machine set spec.
+
+### Integration with machine health check
+
+As the control plane machine set can now create replacement machines, control plane machines may be targeted by a
+machine health check.
```
jeana-redhat, line 85: > may be targeted

What's going to happen by default for clusters using the CPMSO? Is it up to the user to decide if they use health checks for control plane machines? Do we have a stance on whether or not it's a good idea?

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+The control plane machine set is currently only supported on Amazon Web Services (AWS) and Microsoft Azure.
+Other platforms may be supported at a later date.
+
+Google Cloud Platform and OpenStack are planned for inclusion from OpenShift version 4.13 onwards.
+
+## Installation
+
+Installation instructions for control plane machine set can be found in the [installation docs](./installation.md).
+
+## Glossary
```
jeana-redhat, line 119: Ty, very helpful for me

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+          machine.openshift.io/cluster-api-machine-role: master
+          machine.openshift.io/cluster-api-machine-type: master
+          machine.openshift.io/cluster-api-cluster: <cluster-id> [5]
+      spec:
+        providerSpec:
+          value:
+            <platform provider spec> [6]
+```
+
+1. Replicas is 3 in most cases. Support exceptions may allow this to be 5 replicas in certain circumstances.
```
jeana-redhat, line 47: Do we have any idea how common a 5-replica control plane configuration is?

More generally, I'm not familiar with how much we talk about special circumstances due to support exceptions in user docs. I could see it either being helpful or causing complications. I will look into it in the doc set.

docs/user/update-strategies.md
```diff
@@ -0,0 +1,103 @@
...
+
+  D --> |No| CRM
+  D --> |Yes| End
+```
+
+## OnDelete
+
+The `OnDelete` strategy is similar in concept to a statefulset on-delete strategy. It is intended as a manually
+triggered way to replace the machines when there is some update to the configuration.
+This means it can be used to slowly roll out update to the control plane machines.
```
jeana-redhat, line 68:
```suggestion
This means it can be used to slowly roll out updates to the control plane machines.
```

docs/user/update-strategies.md
```diff
@@ -0,0 +1,103 @@
...
+
+Where the `OnDelete` strategy differs from the `RollingUpdate` strategy is that it will not automatically create
```
jeana-redhat, line 70: How does this mesh with having a health check configured? The results of the health check become purely informational for the user to act on when/as desired?

jeana-redhat (commented):
Very well explained. Added a few notes just for myself to cross reference, and added a couple questions where I want to make sure I understand right.

docs/user/README.md
jeana-redhat: Talked to Eric about this, prod docs will just talk about the normal case (3 replicas)

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+Horizontal scaling is not currently supported and so this field is currently immutable.
+This may change in a future release.
+2. The strategy defaults to `RollingUpdate`. `OnDelete` is also supported.
+3. The ControlPlaneMachineSet spreads Machines across multiple failure domains where possible.
+This field must be set to the platform name.
+4. The failure domains vary on their configuration per platform, see below for how to configure a failure domain on
+each platform.
+5. The cluster ID is required here. You should be able to find this label on existing Machines in the cluster.
```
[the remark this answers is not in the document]
JoelSpeed: Oh nice, didn't realise we had that! I'll add in

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+6. The provider spec must match that of the Control Plane Machines created by the installer, except, you can omit any
+field set in the failure domains.
+
+## Features
+
+### Spreading machines across Failure Domains
+
+The control plane machine set can balance machines across multiple failure domains. This provides high availability and
+fault tolerance within the control plane when issues arise within the infrastructure provider.
```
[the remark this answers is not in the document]
JoelSpeed: Yes, it's likely that the control plane is already spread across failure domains implicitly by the installer, this makes it explicit and gives them the ability to change that set of failure domains should they so desire to

docs/user/README.md
JoelSpeed: > Is it up to the user to decide if they use health checks for control plane machines?

Yes, users will need to decide if they want to use it or not.

> Do we have a stance on whether or not it's a good idea?

I would ask support what they think, technically there should be no real issue as the CPMS should start recovering the cluster if any MHC action is taken, but they would definitely want it to be configured strictly with the recommendations I outlined here

docs/user/README.md
```diff
@@ -0,0 +1,146 @@
...
+control plane machine set spec.
+
+### Integration with machine health check
+
+As the control plane machine set can now create replacement machines, control plane machines may be targeted by a
+machine health check.
+It is advised to set the machine health check `maxUnhealthy` value to `1` when configuring machine health checks with
+control plane machines as a target.
+Any machine health check targeting control plane machines should target only control plane machines and should not be
+used to also cover worker machines.
```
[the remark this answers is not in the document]
JoelSpeed: Use the `machine.openshift.io/cluster-api-machine-role` label to distinguish between them, this will be configured in the selector on the MHC

docs/user/README.md
JoelSpeed: Added a little extra colour in the first paragraph of this block, let me know if that helps

docs/user/installation.md
JoelSpeed: I think this was just an omission on my part, will add a short text here now

docs/user/update-strategies.md
JoelSpeed: The health check, when failing, will mark the machine for deletion, triggering the replacement to be created. The MHC is a way to automate deleting instances

docs/user/README.md
lobziik: Yup, to me looks less cryptic now. Thanks!

JoelSpeed:
/approve

Spoke with the other reviewers, ok to merge, will follow up if there's any additional feedback

PR #93, opened 2022-09-07
Ensure machines that were previously ready are still considered ready

JoelSpeed:
A Machine that has not yet been removed from the cluster (ie it has a valid noderef) is typically still ready unless it is failed.
For example, when a Machine is deleted, but the new Machine has not yet come up to replace it, the Machine is still ready as no actions are taken until the etcd operator removes its lifecycle hook.

Currently we go degraded based on a lack of ready machines returned by the machine provider, in the case where a user deletes all machines, this would degrade the operator and we wouldn't automatically replace the deleted machines.

By accepting that a deleted, but previously ready machine is still ready, we can continue operating in this state.


elmiko (commented):
this is a good improvement
/lgtm

PR #98, opened 2022-09-21
OCPBUGS-996: Add logic to handle extra updated machines in a single index + minor fixes

damdo:
This PR adds logic to handle a corner case where there is more than one machine that is Updated (Ready and Up-to-date) in a single index of the Control Plane Machine Set.

As discussed in the [linked issue](https://issues.redhat.com/browse/OCPBUGS-996), the behaviour, covered by the logic introduced in the PR (last two commits) should be the following:
- at cluster state validation stage, check for every index of the CPMS if there is more than one Updated Machine. If that's the case, carry on if the update strategy is `RollingUpdate` or `Recreate`; if instead the strategy is `OnDelete` log an error and put the operator in degraded state.
- if in `RollingUpdate` or `Recreate` in the respective reconcile logics handle the Updated machine in excess by deleting the oldest between the two.

Along the line I found a couple of minor bugs to do with:
- log entries not being checked in the unit tests. I had them checked and fixed their entries to match the expected ones.
- for `fetchControlPlaneNodes` I changed the returning map (that can't be sorted by key) to be a slice, which preserves append order, to have a consistent ordering for the elements in testing.

I suggest reviewing it by commit.

### DIFF
--- a/pkg/controllers/controlplanemachineset/consts.go
+++ b/pkg/controllers/controlplanemachineset/consts.go
@@ -127,6 +127,11 @@ const (
 	// This will typically occur when an old replica has not yet been removed.
 	reasonExcessReplicas = "ExcessReplicas"
 
+	// reasonExcessUpdatedReplicas denotes that the ControlPlaneMachineSet has a more than expected number
+	// of ready and up-to-date replicas.
+	// This will typically occur when extra replica(s) for an index have been created.
+	reasonExcessUpdatedReplicas = "ExcessUpdatedReplicas"
+
 	// reasonNeedsUpdateReplicas denotes that the ControlPlaneMachineSet has identified
 	// replicas under its management that are currently in need of an update.
 	reasonNeedsUpdateReplicas = "NeedsUpdateReplicas"
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -73,6 +73,9 @@ var (
 
 	// errFoundExcessiveIndexes is used to inform users that an excessive number of indexes has been found.
 	errFoundExcessiveIndexes = errors.New("found an excessive number of indexes for the control plane machine set")
+
+	// errFoundExcessiveUpdatedReplicas is used to inform users that an excessive number of updated machines has been found for a single index.
+	errFoundExcessiveUpdatedReplicas = errors.New("found an excessive number of updated machines for a single index")
 )
 
 // ControlPlaneMachineSetReconciler reconciles a ControlPlaneMachineSet object.
@@ -454,6 +457,11 @@ func (r *ControlPlaneMachineSetReconciler) validateClusterState(ctx context.Cont
 		return nil
 	}
 
+	// Check that the number of Updated (Ready and with Up-to-date spec) Machines in an index is valid.
+	if ok := r.checkValidNumerOfUpdatedMachinesPerIndex(logger, cpms, sortedIndexedMs); !ok {
+		return nil
+	}
+
 	// Check that no replacement machines (one that doesn't need update but has an equivalent in the index that needs update)
 	// have an error.
 	if ok := r.checkNoErrorForReplacements(logger, cpms, sortedIndexedMs); !ok {
@@ -655,6 +663,59 @@ func (r *ControlPlaneMachineSetReconciler) checkCorrectNumberOfIndexes(logger lo
 		})
 
 		return false
+	}
+
+	return true
+}
+
+// checkValidNumerOfUpdatedMachinesPerIndex checks that the number of updated machines in an index is valid.
+func (r *ControlPlaneMachineSetReconciler) checkValidNumerOfUpdatedMachinesPerIndex(logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, sortedIndexedMs [][]machineproviders.MachineInfo) bool {
+	for _, index := range sortedIndexedMs {
+		updatedMachinesCount := len(updatedMachines(index))
+
+		switch {
+		case updatedMachinesCount <= 1:
+			// Valid numer of Updated Machines. The cluster state is valid.
+		case updatedMachinesCount > 1:
+			switch cpms.Spec.Strategy.Type {
+			case machinev1.RollingUpdate:
+				// Even though there is an excess number of Updated Machines,
+				// with the RollingUpdate update strategy we consider this a valid state for the cluster.
+				// We carry on and let the RollingUpdate reconciliation handle the excess in Updated Machines.
+			case machinev1.Recreate:
+				// Even though there is an excess number of Updated Machines,
+				// with the Recreate update strategy we consider this a valid state for the cluster.
+				// We carry on and let the Recreate reconciliation handle the excess in Updated Machines.
+			case machinev1.OnDelete:
+				// Too many Updated Machines. With OnDelete update strategy,
+				// if there are an excess number of Updated Machines
+				// we set the operator to degraded and ask the user for manual intervention,
+				// to remove the excess replicas.
+				excessiveUpdatedReplicas := updatedMachinesCount - 1
+
+				logger.Error(
+					fmt.Errorf("%w: %d updated replica(s) are in excess for index %d",
+						errFoundExcessiveUpdatedReplicas, excessiveUpdatedReplicas, index[0].Index),
+					"Observed an excessive number of updated replica(s) for a single index",
+					"excessUpdatedReplicas", excessiveUpdatedReplicas,
+				)
+
+				meta.SetStatusCondition(&cpms.Status.Conditions, metav1.Condition{
+					Type:   conditionProgressing,
+					Status: metav1.ConditionFalse,
+					Reason: reasonOperatorDegraded,
+				})
+
+				meta.SetStatusCondition(&cpms.Status.Conditions, metav1.Condition{
+					Type:    conditionDegraded,
+					Status:  metav1.ConditionTrue,
+					Reason:  reasonExcessUpdatedReplicas,
+					Message: fmt.Sprintf("Observed %d updated machine(s) in excess for index %d", excessiveUpdatedReplicas, index[0].Index),
+				})
+
+				return false
+			}
+		}
 	}
 
 	return true
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -703,9 +703,10 @@ func (r *ControlPlaneMachineSetReconciler) checkNoErrorForReplacements(logger lo
 	return true
 }
 
-// fetchControlPlaneNodes fetches a map of unique nodes that have the "control-plane" (and/or legacy "master") labels.
-func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Context) (map[string]corev1.Node, error) {
-	cpmsNodes := make(map[string]corev1.Node)
+// fetchControlPlaneNodes fetches a sorted list of unique nodes that have the "control-plane" (and/or legacy "master") labels.
+func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Context) ([]corev1.Node, error) {
+	cpmsNodesLookup := make(map[string]struct{})
+	sortedCpmsNodes := []corev1.Node{}
 
 	for _, label := range []string{masterNodeRoleLabel, controlPlaneNodeRoleLabel} {
 		nodesList := &corev1.NodeList{}
@@ -716,9 +717,13 @@ func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Co
 		}
 
 		for _, n := range nodesList.Items {
-			cpmsNodes[n.ObjectMeta.Name] = n
+			if _, exists := cpmsNodesLookup[n.ObjectMeta.Name]; !exists {
+				cpmsNodesLookup[n.ObjectMeta.Name] = struct{}{}
+
+				sortedCpmsNodes = append(sortedCpmsNodes, n)
+			}
 		}
 	}
 
-	return cpmsNodes, nil
+	return sortedCpmsNodes, nil
 }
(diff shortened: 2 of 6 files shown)


pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -716,9 +778,13 @@ func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Co
 		}
 
 		for _, n := range nodesList.Items {
-			cpmsNodes[n.ObjectMeta.Name] = n
+			if _, exists := cpmsNodesLookup[n.ObjectMeta.Name]; !exists {
```
elmiko, line 781: what is the implication of seeing multiple records with the same name?

(i wouldn't expect this to happen, but i want to understand the logic here since we are talking about situations where there are too many machines)

elmiko (commented):
this generally makes sense to me, i have a question about one part.

elmiko:
one more thing occurred to me, do we need an update to the `reconcileOnDeleteUpdate` function?

pkg/controllers/controlplanemachineset/controller.go
damdo: Good question. We might see duplicated nodes because we are filtering for nodes that have either the old or the new label that indicates a node being a control plane (using `masterNodeRoleLabel`, `controlPlaneNodeRoleLabel`).

Since later we map nodes to machines, If a node for example has both labels (imagine in a transition situation), and shows up twice we might end up in a confusing situation, hence the deduping with a lookup set.

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -703,9 +703,10 @@ func (r *ControlPlaneMachineSetReconciler) checkNoErrorForReplacements(logger lo
 	return true
 }
 
-// fetchControlPlaneNodes fetches a map of unique nodes that have the "control-plane" (and/or legacy "master") labels.
-func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Context) (map[string]corev1.Node, error) {
-	cpmsNodes := make(map[string]corev1.Node)
+// fetchControlPlaneNodes fetches a sorted list of unique nodes that have the "control-plane" (and/or legacy "master") labels.
+func (r *ControlPlaneMachineSetReconciler) fetchControlPlaneNodes(ctx context.Context) ([]corev1.Node, error) {
+	cpmsNodesLookup := make(map[string]struct{})
+	sortedCpmsNodes := []corev1.Node{}
```
JoelSpeed, line 770: Have you considered using a `sets.String` here? https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String

It will allow you to iterate through the list and `Insert` without checking if it's already there, then `stringSet.List()` returns a sorted list for you, perhaps that could make this logic a little nicer?

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -919,7 +919,7 @@ var _ = Describe("validateClusterState", func() {
 		}
 
 		Expect(cpms.Status.Conditions).To(test.MatchConditions(in.expectedConditions))
-		Expect(in.expectedLogs).To(ConsistOf(in.expectedLogs))
+		Expect(logger.Entries()).To(ConsistOf(in.expectedLogs))
```
JoelSpeed, line 923: Doh, who did this? 🙈

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -660,6 +668,59 @@ func (r *ControlPlaneMachineSetReconciler) checkCorrectNumberOfIndexes(logger lo
...
 }
 
+// checkValidNumerOfUpdatedMachinesPerIndex checks that the number of updated machines in an index is valid.
+func (r *ControlPlaneMachineSetReconciler) checkValidNumerOfUpdatedMachinesPerIndex(logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, sortedIndexedMs [][]machineproviders.MachineInfo) bool {
+	for _, index := range sortedIndexedMs {
+		updatedMachinesCount := len(updatedMachines(index))
+
+		switch {
+		case updatedMachinesCount == 0 || updatedMachinesCount == 1:
+			// Valid numer of Updated Machines. The cluster state is valid.
```
JoelSpeed, line 678: If you return true here, I think you can remove the final one at the end and it's obvious that this is a happy state.

That or rename this to default as if it's `<=1` it would fall into the default based on the other cases

damdo:
> one more thing occurred to me, do we need an update to the reconcileOnDeleteUpdate function?

Do you mean if we need to add [this same check](https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/98/files#diff-649f5a2c6b4422e2b13f37659b81ad6fbf227565fcd3a37573cad851a30e266eR320-R326) we added for the `reconcileMachineRollingUpdate` `reconcileOnDeleteUpdate`? 

We don't and shouldn't as this check will not be necessary in the OnDelete case since [the operator will go degraded if we end up in this situation and the strategy is set to OnDelete](https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/98/files#diff-7c43826e415d01fd024a3ea41524b078e6b4e2ccb9df2b5972e4f13489d0a9dcR679-R716).

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -1171,6 +1172,71 @@ var _ = Describe("validateClusterState", func() {
...
 		}),
+		Entry("with multiple updated machines in a single index and RollingUpdate strategy", validateClusterTableInput{
+			cpmsBuilder: cpmsBuilder.WithConditions([]metav1.Condition{
+				degradedConditionBuilder.WithStatus(metav1.ConditionFalse).Build(),
+				progressingConditionBuilder.WithStatus(metav1.ConditionFalse).Build(),
+			}).WithReplicas(3),
+			machineInfos: map[int32][]machineproviders.MachineInfo{
+				0: {
+					updatedMachineBuilder.WithIndex(0).WithMachineName("machine-0").WithNodeName("master-0").Build(),
+					updatedMachineBuilder.WithIndex(0).WithMachineName("machine-0").WithNodeName("master-0").Build(),
```
JoelSpeed, line 1183: This machine should have `-replacement` in the middle of the machine and node names, make sure the names are unique

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -1171,6 +1172,71 @@ var _ = Describe("validateClusterState", func() {
...
+		}),
+		Entry("with multiple updated machines in a single index and OnDelete strategy", validateClusterTableInput{
+			cpmsBuilder: cpmsBuilder.WithConditions([]metav1.Condition{
+				degradedConditionBuilder.WithStatus(metav1.ConditionFalse).Build(),
+				progressingConditionBuilder.WithStatus(metav1.ConditionFalse).Build(),
+			}).WithReplicas(3).WithStrategyType(machinev1.OnDelete),
+			machineInfos: map[int32][]machineproviders.MachineInfo{
+				0: {
+					updatedMachineBuilder.WithIndex(0).WithMachineName("machine-0").WithNodeName("master-0").Build(),
+					updatedMachineBuilder.WithIndex(0).WithMachineName("machine-0").WithNodeName("master-0").Build(),
```
JoelSpeed, line 1211: This machine should have `-replacement` in the middle of the machine and node names, make sure the names are unique

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -1171,6 +1172,71 @@ var _ = Describe("validateClusterState", func() {
...
+				},
+				1: {updatedMachineBuilder.WithIndex(1).WithMachineName("machine-1").WithNodeName("master-1").Build()},
+				2: {updatedMachineBuilder.WithIndex(2).WithMachineName("machine-2").WithNodeName("master-2").Build()},
+			},
+			nodes: []*corev1.Node{
+				masterNodeBuilder.WithName("master-0").Build(),
```
JoelSpeed, line 1189: Where's the replacement node in this list?

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -1171,6 +1172,71 @@ var _ = Describe("validateClusterState", func() {
...
```
JoelSpeed, line 1217: Where's the replacement node in this list?

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -1765,4 +1766,38 @@ var _ = Describe("utils tests", func() {
...
+		func(input []machineproviders.MachineInfo, expected []machineproviders.MachineInfo) {
+			output := sortMachineInfoByCreationTimestamp(input)
+			Expect(output).To(Equal(expected))
+		},
+		Entry("when MachineInfo is not sorted by CreationTimestamp",
+			[]machineproviders.MachineInfo{
+				updatedMachineBuilder.WithIndex(0).WithMachineName("machine-newer-0").WithNodeName("node-0").
+					WithMachineCreationTimestamp(metav1.NewTime(time.Date(2022, 01, 01, 01, 02, 01, 01, time.UTC))).Build(),
+				updatedMachineBuilder.WithIndex(0).WithMachineName("machine-older-0").WithNodeName("node-0").
+					WithMachineCreationTimestamp(metav1.NewTime(time.Date(2022, 01, 01, 01, 01, 01, 01, time.UTC))).Build(),
```
JoelSpeed, line 1852: Can we vary the last numeric value in the list, or the month or something? It took me a while to notice the middle (hour) value was varied

pkg/test/resourcebuilder/machine_info.go
```diff
@@ -82,6 +84,12 @@ func (m MachineInfoBuilder) Build() machineproviders.MachineInfo {
 	return info
 }
 
+// WithMachineCreationTimestamp sets the machine deletion timestamp for the machineinfo builder.
+func (m MachineInfoBuilder) WithMachineCreationTimestamp(deletion metav1.Time) MachineInfoBuilder {
+	m.machineCreationtimestamp = deletion
```
JoelSpeed, line 89: You need to replace the word `deletion` with `creation` here

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -779,6 +779,79 @@ var _ = Describe("reconcileMachineUpdates", func() {
 					}
 				},
 			}),
+			Entry("with an extra updated machine in a single index", rollingUpdateTableInput{
+				cpmsBuilder: cpmsBuilder.WithReplicas(3),
+				machineInfos: map[int32][]machineproviders.MachineInfo{
+					0: {
+						updatedMachineBuilder.WithIndex(0).WithMachineName("machine-older-extra-0").WithNodeName("node-0").Build(),
```
JoelSpeed, line 786: Vary the node name

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -779,6 +779,79 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+							Message: removingOldMachine,
+						},
+					}
+				},
+			}),
+			Entry("with extra updated machines in multiple indexes", rollingUpdateTableInput{
+				cpmsBuilder: cpmsBuilder.WithReplicas(3),
+				machineInfos: map[int32][]machineproviders.MachineInfo{
+					0: {
+						updatedMachineBuilder.WithIndex(0).WithMachineName("machine-older-extra-0").WithNodeName("node-0").Build(),
```
JoelSpeed, line 815: Vary the node name

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -779,6 +779,79 @@ var _ = Describe("reconcileMachineUpdates", func() {
...
+						updatedMachineBuilder.WithIndex(0).WithMachineName("machine-0").WithNodeName("node-0").Build(),
+					},
+					1: {
+						updatedMachineBuilder.WithIndex(1).WithMachineName("machine-older-extra-1").WithNodeName("node-1").Build(),
```
JoelSpeed, line 819: Vary the node name

pkg/controllers/controlplanemachineset/controller_test.go
damdo: Hehe :) no worries, it happens.

pkg/controllers/controlplanemachineset/controller.go
damdo: We can't return true because we are looping through all the indexes and we should return happy only if all of them don't have an extra updated machine in every index. I can make `updatedMachinesCount == 0 || updatedMachinesCount == 1` -> `updatedMachinesCount <= 1`.

pkg/controllers/controlplanemachineset/controller_test.go
damdo: Oops, thanks!

pkg/controllers/controlplanemachineset/updates_test.go
damdo: Fair :D

pkg/controllers/controlplanemachineset/consts.go
```diff
@@ -127,6 +127,11 @@ const (
 	// This will typically occur when an old replica has not yet been removed.
 	reasonExcessReplicas = "ExcessReplicas"
 
+	// reasonExcessUpdatedReplicas denotes that the ControlPlaneMachineSet has a more than expected number
+	// of ready and up-to-date replicas.
+	// This will typically occur when extra replica(s) for an index has been created.
```
JoelSpeed, line 132: Nit, `an extra replica has`, vs, `extra replicas have`, given we are missing `an` I think `have`
```suggestion
	// This will typically occur when extra replica(s) for an index have been created.
```

JoelSpeed (commented):
/approve
/lgtm

One wording nit but not blocking

JoelSpeed (commented):
/lgtm
/hold cancel

PR #101, opened 2022-09-27
OCPBUGS-996: Reverts "Add logic to handle extra updated machines in a single index + minor fixes"

stbenjam:
Reverts openshift/cluster-control-plane-machine-set-operator#98 ; tracked by OCPBUGS-1752

Per TRT policy, we are reverting this breaking change to get ci and/or nightly payloads flowing again.  Since this PR merged, aws-sdn-serial has been failing on control plate-related tasks, such as the etcd scaling test which adds an additional control plane node.  Looking at the run history on the PR, the test was also failing in presubmits.  Prior to this merging, this test passed at 95%, and has since dropped to under 60%. https://amd64.ocp.releases.ci.openshift.org/releasestream/4.12.0-0.ci/release/4.12.0-0.ci-2022-09-26-133725 is the first payload that failed, and this looks like the most obvious change to me.

To unrevert this revert, revert this PR, and layer an additional separate commit on top that addresses the problem.  Ensure e2e-aws-serial passes.

CC: @damdo, @JoelSpeed

### DIFF
--- a/pkg/controllers/controlplanemachineset/consts.go
+++ b/pkg/controllers/controlplanemachineset/consts.go
@@ -127,11 +127,6 @@ const (
 	// This will typically occur when an old replica has not yet been removed.
 	reasonExcessReplicas = "ExcessReplicas"
 
-	// reasonExcessUpdatedReplicas denotes that the ControlPlaneMachineSet has a more than expected number
-	// of ready and up-to-date replicas.
-	// This will typically occur when extra replica(s) for an index have been created.
-	reasonExcessUpdatedReplicas = "ExcessUpdatedReplicas"
-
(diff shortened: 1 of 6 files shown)


stbenjam:
The only commit after this is 925943433d4c224465ae605e8eb6550926dc0dd4, which was a docs-only change, so reverting this restores a previously tested HEAD.  @JoelSpeed also confirmed this is likely the cause, so I am approving and overriding to unblock payloads

/label approved
/override ci/prow/e2e-aws-ovn

stbenjam:
/retitle Reverts "Add logic to handle extra updated machines in a single index + minor fixes"

PR #103, opened 2022-09-28
Ensure that an inactive CPMS does not cause changes to the cluster

JoelSpeed:
We want to add an active/inactive mode to the CPMS.

When inactive, the CPMS will calculate the status and do nothing else. It will set the cluster operator available in case there are errors that a user needs to fix.

Once the user marks the CPMS active, this will put the CPMS into operational mode and any issues on the CPMS status will reflect into the cluster operator, equally, the CPMS will start to operate.

We expect users to see the inactive CPMS and check the status/resolve any issues before marking it active.


pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -162,6 +210,302 @@ var _ = Describe("With a running controller", func() {
...
+			It("should add an owner reference to each machine", func() {
+				Eventually(komega.ObjectList(&machinev1beta1.MachineList{})).Should(HaveField("Items", Not(ContainElement(HaveField("ObjectMeta.OwnerReferences", BeEmpty())))), "No machine should not have an owner reference")
+			})
+
+			It("should create a replacement for the machine", func() {
+				Eventually(komega.ObjectList(&machinev1beta1.MachineList{})).Should(HaveField("Items", HaveLen(4)))
+			})
+		})
+
+		Context("with no running machines", func() {
```
elmiko, line 290: does this represent a situation where all 3 machines are in the process of becoming nodes?

elmiko (commented):
this generally makes sense, i have a couple clarification questions and i'm also curious if we are co-mingling the openstack removal with the active/inactive update?

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -162,6 +210,302 @@ var _ = Describe("With a running controller", func() {
...
+					WithType(conditionDegraded).
+					WithStatus(metav1.ConditionTrue).
+					WithReason(reasonNoReadyMachines).
+					WithMessage("No ready control plane machines found").
+					Build()
+
+				Eventually(komega.Object(cpms)).Should(HaveField("Status.Conditions", ContainElement(test.MatchCondition(degradedCondition))))
+			})
+
+			It("should the cluster operator as available", func() {
```
damdo, line 482:
```suggestion
			It("should set the cluster operator as available", func() {
```

pkg/controllers/controlplanemachineset/controller_test.go
JoelSpeed: Or a situation where you have 3 broken machines that will never become nodes. It's a scenario I know makes the operator go degraded so is easy to use to prove the cluster operator conditions sync

elmiko (commented):
thanks for the answer/updates
/lgtm

damdo (commented):
/lgtm

PR #104, opened 2022-09-28
OCPCLOUD-1673: Add Control Plane Machine Set generator controller

damdo:
We need a way to generate the CPMS object for a cluster based on its current state.

We want a new controller that based on the cluster state generates a CPMS object and creates it against the API of the cluster. The CPMS object that will be created will have the `.spec.state` field set to `Inactive`, meaning that the main CPMS controller won't do any modification to the cluster based on that configuration until the user has reviewed/modified it and activated it by setting it to `Active`.

To catch changes in the cluster state the CPMS generator controller will keep watching the state and perform CPMS object regenerations checking if the object needs updating. If that's the case the generator will delete and recreate the CPMS object. This will constantly happen while the CPMS state is set to `Inactive`. As soon as the user activates it (sets `.spec.state` to `Active`). The generator controller will stop processing and its logic will become a no-op.

For now the generator controller will only be able to generate config for AWS clusters.


manifests/0000_31_control-plane-machine-set-operator_01_rbac.yaml
```diff
@@ -41,6 +40,28 @@ rules:
...
+    resources:
+      - controlplanemachinesets
+    verbs:
+      - get
+      - list
+      - watch
+      - update
+      - patch
+      - create
+      - delete
```
JoelSpeed, line 54: Do we already have a block for controlplanemachinesets? I'd expect so

manifests/0000_31_control-plane-machine-set-operator_01_rbac.yaml
```diff
@@ -41,6 +40,28 @@ rules:
       - update
       - patch
 
+  - apiGroups:
+      - machine.openshift.io
+    resources:
+      - controlplanemachinesets
```
JoelSpeed, line 46: Do we not already have a block for this?

pkg/controllers/controlplanemachinesetgenerator/aws.go
```diff
@@ -0,0 +1,98 @@
...
+	awsPs.Placement.AvailabilityZone = ""
+
+	// Re-encode the ProviderSpec.
+	re, err := rawExtensionFromProviderSpec(platformType, awsPs)
+	if err != nil {
+		return machinev1builder.ControlPlaneMachineSetSpecApplyConfiguration{}, fmt.Errorf("failed to encode machine providerSpec: %w", err)
+	}
+
+	// We always assume 3 replicas here. There might be very rare occasions where the number differs,
+	// but we will let the user change the generated CPMS (before activating it) afterwards if they desire to set that differently.
```
JoelSpeed, line 53: Currently the field is immutable so this comment isn't true, unless there's a way to override the validation, maybe you can try this

pkg/controllers/controlplanemachinesetgenerator/controller.go
```diff
@@ -0,0 +1,327 @@
...
+	// Any ControlPlaneMachineSet not in this namespace should be ignored.
+	Namespace string
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ControlPlaneMachineSetGeneratorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// All predicates are executed before the event handler is called.
+	if err := ctrl.NewControllerManagedBy(mgr).
+		For(&machinev1.ControlPlaneMachineSet{}, builder.WithPredicates(filterControlPlaneMachineSet(r.Namespace))).
+		Owns(&machinev1.ControlPlaneMachineSet{}, builder.WithPredicates(filterControlPlaneMachineSet(r.Namespace))).
```
JoelSpeed, line 88: If you have a `For` we don't need `Owns` too I think

pkg/controllers/controlplanemachinesetgenerator/controller.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	// Compare if the current and the newly generated CPMS spec match.
+	if diff := compareCPMSs(infrastructure.Spec.PlatformSpec.Type, *cpms, *generatedCpms); diff != nil {
+		// The two CPMSs don't match.
+		logger.V(1).WithValues("diff", diff).Info("outdated control plane machine set found")
+
+		// Delete the current and ReCreate
+		// a new one with the newly generated configuration.
+		if err := r.Delete(ctx, cpms); err != nil {
```
JoelSpeed, line 198: In case the cache is stale, do we want to add a delete precondition here? Just as a double check

manifests/0000_31_control-plane-machine-set-operator_01_rbac.yaml
damdo: For the status and the finalizer yes, but I moved this in a separate block to the role more permission over this (+delete, +create) than over the other two.

manifests/0000_31_control-plane-machine-set-operator_01_rbac.yaml
damdo: See https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/104#discussion_r984729875

pkg/controllers/controlplanemachinesetgenerator/controller.go
damdo: 👍

pkg/controllers/controlplanemachinesetgenerator/controller.go
damdo: Good idea! Added.

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/aws.go
```diff
@@ -64,6 +64,10 @@ func (a AWSProviderConfig) Config() machinev1beta1.AWSMachineProviderConfig {
 // It should return an error if the provided RawExtension does not represent
 // an AWSMachineProviderConfig.
 func newAWSProviderConfig(raw *runtime.RawExtension) (ProviderConfig, error) {
+	if raw == nil {
+		return nil, errNilProviderSpec
+	}
+
```
damdo, line 70: This one fixes a nil pointer dereference issue I found in the aws machineprovider. Unrelated to the controller but found while writing it.

pkg/controllers/controlplanemachinesetgenerator/aws.go
damdo: We chatted offline and found a different workaround. I modified that comment to reflect it.

pkg/controllers/controlplanemachinesetgenerator/controller_test.go
```diff
@@ -0,0 +1,592 @@
...
+				By("Starting the manager")
+				mgrCancel, mgrDone = startManager(&mgr)
+			})
+
+			AfterEach(func() {
+				By("Stopping the manager")
+				stopManager()
+			})
+
+			It("should recreate CPMS with the provider spec matching the youngest machine provider spec", func() {
```
JoelSpeed, line 495: Maybe check the UID of the CPMS changed to make sure it was a delete and recreate?

pkg/controllers/controlplanemachinesetgenerator/controller_test.go
```diff
@@ -0,0 +1,592 @@
...
+				machineProviderSpec, err := extractAWSProviderSpec(machine2.Spec.ProviderSpec)
+				Expect(err).To(BeNil())
+
+				// Remove from the machine Provider Spec the fields that won't be
+				// present on the Control Plane Machine Set Provider Spec.
+				machineProviderSpec.Subnet = machinev1beta1.AWSResourceReference{}
+				machineProviderSpec.Placement.AvailabilityZone = ""
+
+				Eventually(func() *machinev1beta1.AWSMachineProviderConfig {
+					Eventually(komega.Get(cpms)).Should(Succeed())
```
JoelSpeed, line 507: Don't use an eventually within an eventually, just get the object and return an error if there's a failure.

If I were writing this I would probably use komega to get the object and write a tranform to extract the providerspec

JoelSpeed (commented):
A few comments in here cover the whole of the codebase changes here, for example, please don't use CPMS in function names, spell it out, and make sure all error returns are wrapped

pkg/controllers/controlplanemachinesetgenerator/utils.go
```diff
@@ -0,0 +1,210 @@
...
+// sortMachinesByCreationTime sorts a slice of Machines by CreationTime, Name (descending).
+func sortMachinesByCreationTime(machines []machinev1beta1.Machine) []machinev1beta1.Machine {
+	// Sort in inverse order so that the newest one is first.
+	sort.Slice(machines, func(i, j int) bool {
+		first, second := machines[i].CreationTimestamp, machines[j].CreationTimestamp
+		if first != second {
+			return second.Before(&first)
+		}
+
+		return strings.Compare(machines[i].Name, machines[j].Name) == 1
```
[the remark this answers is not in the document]
damdo: Yeah I was unsure on the best approach but the one you suggest might be clearer. Will update.

pkg/controllers/controlplanemachinesetgenerator/utils.go
```diff
@@ -0,0 +1,210 @@
...
+	return diff, nil
+}
+
+// mergeMachineSlices merges two machine slices into one, removing duplicates.
+func mergeMachineSlices(a []machinev1beta1.Machine, b []machinev1beta1.Machine) []machinev1beta1.Machine {
+	combined := []machinev1beta1.Machine{}
+	combined = append(combined, a...)
+	combined = append(combined, b...)
+
+	allKeys := make(map[string]bool)
```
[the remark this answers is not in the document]
damdo: Yeah, I use this sometimes when I want very efficient stuff. Ok, updated.

pkg/controllers/controlplanemachinesetgenerator/aws.go
```diff
@@ -0,0 +1,143 @@
...
+	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	machinev1builder "github.com/openshift/client-go/machine/applyconfigurations/machine/v1"
+	machinev1beta1builder "github.com/openshift/client-go/machine/applyconfigurations/machine/v1beta1"
+)
+
+const (
+	// replicas is the number of Control Plane Machine Set indexes.
+	// We always assume 3 replicas here. There might be very rare occasions where the number differs,
+	// but that's an exeption that must be manually forced on the applied CPMS object.
```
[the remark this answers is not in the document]
damdo: We discussed this online and found there was a misunderstanding. We'll keep it this way (but fix the typo).

pkg/controllers/controlplanemachinesetgenerator/aws.go
```diff
@@ -0,0 +1,143 @@
...
+// buildCPMSAWSMachineSpec builds an AWS flavored MachineSpec for the CPMS.
+func buildCPMSAWSMachineSpec(machines []machinev1beta1.Machine) (*machinev1beta1builder.MachineSpecApplyConfiguration, error) {
+	// Take the Provider Spec of the first in the machines slice
+	// as a the one to be put on the CPMS spec.
+	// Since the `machines` slice is sorted by descending creation time
+	// we are guaranteed to get the newest Provider Spec of a machine.
+	// This is done so that if there are control plane machines with differing
+	// Provider Specs, we will use the most recent one. This is an attempt to try and inferr
+	// the spec that the user might want to choose among the different ones found in the cluster.
+	referenceProviderSpec := machines[0].Spec.ProviderSpec
```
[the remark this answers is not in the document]
damdo: We do an early check on that and don't support clusters when we detect less than 2 control plane machines.
https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/104/files#diff-8e057fdd6e8091414f1b0673b789240eef25ca48e35f13c3669f945f80fcc205R334-R343

pkg/controllers/controlplanemachinesetgenerator/controller.go
```diff
@@ -0,0 +1,343 @@
...
+	// Any ControlPlaneMachineSet not in this namespace should be ignored.
+	Namespace string
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ControlPlaneMachineSetGeneratorReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// All predicates are executed before the event handler is called.
+	if err := ctrl.NewControllerManagedBy(mgr).
+		For(&machinev1.ControlPlaneMachineSet{}, builder.WithPredicates(filterControlPlaneMachineSet(r.Namespace))).
+		Watches(&source.Kind{Type: &machinev1beta1.Machine{}}, handler.EnqueueRequestsFromMapFunc(machineToControlPlaneMachineSet(r.Namespace))).
```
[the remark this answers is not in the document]
damdo: Rebased, extracted moved all the watch filters in the util pkg, and used a consistent approach across both controllers.

pkg/util/watch_filters.go
```diff
@@ -40,23 +40,26 @@ const (
...
+	machineControlPlaneTypeLabelName = "control-plane"
 )
 
-// objToControlPlaneMachineSet maps any object to the control plane machine set singleton
+// ObjToControlPlaneMachineSet maps any object to the control plane machine set singleton
 // in the namespace provided.
-func objToControlPlaneMachineSet(namespace string) func(client.Object) []reconcile.Request {
+func ObjToControlPlaneMachineSet(controlPlaneMachineSetName, namespace string) func(client.Object) []reconcile.Request {
 	return func(obj client.Object) []reconcile.Request {
 		klog.V(4).Infof("reconcile triggered by object (%T) %s/%s", obj, obj.GetNamespace(), obj.GetName())
```
JoelSpeed, line 52: I've got a PR updating this line, will need to co-ordinate changes here

pkg/controllers/controlplanemachinesetgenerator/aws.go
```diff
@@ -0,0 +1,143 @@
...
+// generateControlPlaneMachineSetAWSSpec generates an AWS flavored ControlPlaneMachineSet Spec.
+func generateControlPlaneMachineSetAWSSpec(machines []machinev1beta1.Machine, machineSets []machinev1beta1.MachineSet) (machinev1builder.ControlPlaneMachineSetSpecApplyConfiguration, error) {
+	awsFailureDomains, err := buildAWSFailureDomains(machineSets)
+	if err != nil {
+		return machinev1builder.ControlPlaneMachineSetSpecApplyConfiguration{}, fmt.Errorf("failed to build aws failure domains config: %w", err)
+	}
+
+	controlPlaneMachineSetMachineSpec, err := buildControlPlaneMachineSetAWSMachineSpec(machines)
+	if err != nil {
+		return machinev1builder.ControlPlaneMachineSetSpecApplyConfiguration{}, fmt.Errorf("failed to build aws failure domains config: %w", err)
```
JoelSpeed, line 45: This message doesn't look correct given the function called above

pkg/util/watch_filters.go
damdo: Ok the first that merges is fine, the other fixes the conflict. WDYT?

JoelSpeed:
Yep, changes look good, lets get this squashed

damdo:
Squashed 👍

PR #105, opened 2022-09-29
OCPBUGS-1411: Bump k8s dependencies to 1.25

lobziik:


damdo (commented):
Provided tests pass:
/lgtm

damdo (commented):
/lgtm

.golangci.yaml
```diff
@@ -4,16 +4,13 @@ linters:
   # https://golangci-lint.run/usage/linters/#enabled-by-default-linters
   enable:
     # Default linters
-    - deadcode
```
JoelSpeed, line 7: Why are we dropping linters?

JoelSpeed (commented):
/approve
/lgtm

.golangci.yaml
lobziik: It was deprecated and might be covered with the `unused` one.
Some context around: https://github.com/golangci/golangci-lint/pull/3125

PR #106, opened 2022-09-29
OCPBUGS-1348: Fix deleting machine affecting new machine failure domain

RadekManak:
When failure domains are mapped onto indexes, we first try to assign failure domains of machines to indexes and after that we populate empty indexes with remaining failure domains. We iterate over indexes in ascending order to be consistent.

This caused a bug if you deleted a machine on index 0. The first thing that would happen is it would notice the deleting machine on index 0 and assign its failure domain to the index. Even if there were more machines with that failure domain over the maximum allowed.

To fix this, I have removed machines in deleting phase from affecting failure domain mapping.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
@@ -29,6 +29,7 @@ import (
 	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
 
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig"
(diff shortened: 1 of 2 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -123,6 +124,11 @@ func createMachineMapping(ctx context.Context, logger logr.Logger, cl client.Cli
 	indexToMachine := make(map[int32]machinev1beta1.Machine)
 
 	for _, machine := range machineList.Items {
+		// When the machine is in deleting phase,
+		// we should not take it into account when calculating where to place the new machine.
+		if pointer.StringDeref(machine.Status.Phase, "") == "Deleting" {
+			continue
+		}
```
JoelSpeed, line 131: What if the deleting machine is the only machine in the index, are we sure this is what we want?

RadekManak: That's the case in the test. Without this change, the deleting machine gets matched to index 0 then indexes 1 and 2 get populated from base because failure domain "us-east-1a" is fully represented in deleting machine at index 0

With this change, it assigns index 1 to "us-east-1a" because there is a ready machine there. This swaps the failure domain in indexes 0 and 1 causing the candidate for index 0 to be "us-east-1b". Index 0 gets assigned "us-east-1b" because it is empty. Index 2 gets assigned "us-east-1c" because failure domain "us-east-1a" in that index is overrepresented.

This problem occurs only in the OnDelete strategy because the user deleted a machine that was considered healthy. In fact, they were all equally healthy, but the algorithm had to choose which ones are the overrepresented.

RadekManak:
Fixed the linter errors.

JoelSpeed:
Have reviewed the bug and the code again and I can see why this makes sense
/lgtm
/approve

Good work here understanding the issue!

JoelSpeed:

Known OVN failure, sending alerts because of 4 masters

PR #110, opened 2022-10-05
OCPBUGS-1645: Ensure the ControlPlaneMachineSet can handle machines that are not indexed from 0

JoelSpeed:
We expect over time that cluster may have renamed their control plane machines, for example rather than being 0, 1, 2, they may now be , 3, 4, 5 or even 2, 4, 6. Previously the CPMS was not handling this and would degrade if it ever saw non 0, 1, 2 indexes.

This PR updates the logic and adds a number of tests to ensure that the entire controller supports these topologies:
* Tests added for full integration checking that the status is set correctly when observing a machine state with various indexes
* Fixed the machineInfosByIndex so that it only adds empty indexes when there aren't the expected number
* Fixed machine provider so that it maps the indexes based on the preferred indexes gathered from the Machines
* Added tests for various parts of the machine provider to check it works as expected

Have then manually tested this on a cluster and can confirm it now is working as we would expect.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
@@ -190,14 +190,16 @@ func mapIndexesToFailureDomainsForMachines(logger logr.Logger, machineList *mach
 // is randomised by golang).
 func reconcileMappings(logger logr.Logger, base, machines map[int32]failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
 	// Create the initial output mapping based on the machines.
-	// Remove any extra indexes that aren't present in the base mapping.
-	// This might occur if the control plane has been scaled down horizontally.
 	out := copyMapping(machines)
-	removeExtraIndexes(out, base)
 
 	// Create the list of candidates based on the base mapping.
 	// These will be matched against the machine mapping.
 	candidates := copyMapping(base)
+
+	// Ensure that, if the machines aren't indexed from 0, the candidate mapping reflects
+	// the indexes present in the machine as opposed to the base mapping which will always
+	// index from 0.
+	reconcileIndexes(candidates, out)
 
 	// We need to keep track of the indexes we haven't yet matched.
 	// Create a list of unmatched indexes.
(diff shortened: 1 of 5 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -190,15 +190,17 @@ func mapIndexesToFailureDomainsForMachines(logger logr.Logger, machineList *mach
...
-	removeExtraIndexes(out, base)
 
 	// Create the list of candidates based on the base mapping.
 	// These will be matched against the machine mapping.
 	candidates := copyMapping(base)
 
+	// Ensure that, if the machines aren't indexed from 0, the candidate mapping reflects
+	// the indexes present in the machine as opposed to the base mapping which will always
+	// index from 0.
+	reconcileIndexes(candidates, out)
```
elmiko, line 202: could we give `out` a more descriptive name. i like `candidates` but i am a little confused by `out`. maybe we could just follow the `reconcileIndexes` names and call it `preferred` ?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -290,6 +292,28 @@ func handleUnmatchedIndex(logger logr.Logger, idx int32, out, base, candidates m
 	}
 }
 
+// reconcileIndexes tries to adjust the output mapping based on a list of preferred
+// indexes. This is used so that when Machines aren't indexed exactly as in the base
+// mapping, we still respect their original mappings.
+func reconcileIndexes(out, preferred map[int32]failuredomain.FailureDomain) {
```
elmiko, line 298: minor nit, but could we change `out` to `reconciled` or something a little more descriptive?

elmiko (commented):
i think this mostly makes sense to me after a second read through, i have a couple minor suggestions.

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
JoelSpeed: Out was chosen as it's the output of the mapping from this function, preferred wouldn't work here IMO. This map is the final output of the mapping algorithm, I'm not sure what else would make sense

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
JoelSpeed: Yeah I'm not sure I even like the function name here to be honest, I'm adding tests which might help explain what exactly it does, perhaps we can come up with a better name

JoelSpeed:
While testing yesterday I managed to make the balancing algorithm panic. To resolve this, I've added more tests and swapped the machine and base mapping around. Now, the base mapping will always contain at least as many indexes as the machine mapping. This ensures that there's always enough candidates when the balancing algorithm kicks in.

Added a bunch of tests to make sure outputs are stable

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -393,15 +434,28 @@ func countForFailureDomain(mapping map[int32]failuredomain.FailureDomain, target
...
+		return out[i] < out[j]
+	})
+
+	return out
+}
+
+// pop removes the first entry from a slice and returns the remaining slice.
+func pop[V any](in []V) (V, []V) {
+	return in[0], in[1:]
 }
```
damdo, line 459: Generics 🙌

damdo (commented):
Good job getting this sorted. 👍 
/lgtm

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -290,6 +306,31 @@ func handleUnmatchedIndex(logger logr.Logger, idx int32, out, base, candidates m
...
+// [3:a, 4:b, 2:c], note 2 in this case was fixed as there was overlap.
+func reconcileIndexes(reconciled, preferred map[int32]failuredomain.FailureDomain) {
+	nonPreferredIndexes := extraIndexes(reconciled, preferred)
+	nonReconciledIndexes := extraIndexes(preferred, reconciled)
+
+	// For each index that isn't in the preferred list, swap it for one
+	// that isn't in the reconciled list if there are any.
+	for _, idx := range nonPreferredIndexes {
+		if len(nonReconciledIndexes) == 0 {
+			continue
```
damdo, line 323: Would a `break` be more appropriate in this case?

damdo (commented):
/lgtm

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
elmiko: ack, that makes sense. still confusing to me, but not a blocker

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
elmiko: that sounds good

PR #111, opened 2022-10-07
OCPCLOUD-1675: Add support for generic platforms without failure domains

RadekManak:
This PR allows CPMS to manage control plane machines on all platforms that do not require modification of machine spec for failure domains. 
It does so by creating a generic provider config that contains the spec as RawExtension to allow us to compare machine spec to template spec. 
I've also added resourcebuilder for VSphere for use in test as an example of platform without failure domains.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/generic_platform.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/generic_platform.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package providerconfig
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// GenericProviderConfig holds the provider spec for machine on platforms that
+// don't support failure domains and can be handled generically.
+type GenericProviderConfig struct {
+	// We don't know the exact type of the providerSpec, but we need to store it to be able to tell if two specs are identical.
+	providerSpec *runtime.RawExtension
+}
+
+// ExtractFailureDomain extract generic failure domain that contains just the platform type.
+func (g GenericProviderConfig) ExtractFailureDomain() failuredomain.FailureDomain {
+	return failuredomain.NewGenericFailureDomain()
+}
+
+// newGenericProviderConfig creates a generic ProviderConfig that can contain providerSpec for any platform.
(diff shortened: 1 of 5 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/generic_platform.go
```diff
@@ -0,0 +1,47 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package providerconfig
+
+import (
+	v1 "github.com/openshift/api/config/v1"
```
JoelSpeed, line 20: Please import this as `configv1` as per conventions

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig.go
```diff
@@ -225,17 +242,37 @@ func (p providerConfig) GCP() GCPProviderConfig {
...
+		"LibvirtMachineProviderConfig":      configv1.LibvirtPlatformType,
+		"OpenStackMachineProviderSpec":      configv1.OpenStackPlatformType,
+		"VSphereMachineProviderSpec":        configv1.VSpherePlatformType,
+		"OvirtMachineProviderSpec":          configv1.OvirtPlatformType,
+		"IBMCloudMachineProviderSpec":       configv1.IBMCloudPlatformType,
+		"KubevirtMachineProviderSpec":       configv1.KubevirtPlatformType,
+		"EquinixMetalMachineProviderConfig": configv1.EquinixMetalPlatformType,
+		"PowerVSMachineProviderConfig":      configv1.PowerVSPlatformType,
+		"AlibabaCloudMachineProviderConfig": configv1.AlibabaCloudPlatformType,
+		"NutanixMachineProviderConfig":      configv1.NutanixPlatformType,
```
JoelSpeed, line 266: This list is likely to go out of date quite quickly, is there any way we can avoid this? 

What are we using this lookup for exactly?

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig.go
```diff
@@ -225,17 +242,37 @@ func (p providerConfig) GCP() GCPProviderConfig {
...
 	}
 
 	platformType, ok := providerSpecKindToPlatformType[kind]
 
+	// Attempt to operate on unknown platforms. This should work if the platform does not require failure domains support.
+	if !ok && (strings.Contains(kind, "MachineProviderConfig") || strings.Contains(kind, "MachineProviderSpec")) {
```
JoelSpeed, line 272: Why the check for `MachineProviderConfig` or `MachineProviderSpec`?

JoelSpeed (commented):
Looks good, few questions and nits reading through

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig.go
RadekManak: I needed this before I had the default case for unknown platforms. We only need this for platforms that have failure domains.

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig.go
RadekManak: I wanted a sanity check that it is some kind provider spec, but we can assume it is here and handle the error later if it is invalid.

JoelSpeed: Yeah I think that makes sense, lets not bother here and we can handle other errors later

JoelSpeed (commented):
/lgtm
/approve

RadekManak:
Fixed lint

PR #113, opened 2022-10-11
OCPBUGS-1348: Fix machine not being marked as deleting

RadekManak:
In my previous attempt to fix [OCPBUGS-1348](https://issues.redhat.com/browse/OCPBUGS-1348), I relied on the machine having phase set to "Deleting". For this to be set, it requires the machine to be reconciled by MAO. Unfortunately, CPMS can get the machine before it is reconciled into deleting phase.
 I have changed the code to use DeletionTimestamp. It is set by the apiserver when it processes the deletion, so it should be reliably set when CPMS sees the deletion.


RadekManak:
/retest-required
Works on my machine 🤔

JoelSpeed:
Changes seem ok
/lgtm
/approve

Good catch

PR #114, opened 2022-10-12
Update watches to include unowned control plane machines

JoelSpeed:
When using `Owns`, the events from the machine objects are filtered so that we only reconcile them if the Machine itself has a controller owner reference pointing at the CPMS.
This works well when the CPMS is `Active`, but when `Inactive`, it means we don't reconcile on Machine events. That means that we don't update the status of the CPMS if Machines are updated (which is when we want to in an `Inactive` state).

This change updates the watch so that we reconcile any control plane machine event regardless of ownership.
I also added some additional logging so that at level 4, we can see when a reconcile is being triggered and by what, this was helpful for debugging.

### DIFF
--- a/pkg/controllers/controlplanemachineset/watch_filters.go
+++ b/pkg/controllers/controlplanemachineset/watch_filters.go
@@ -17,9 +17,12 @@ limitations under the License.
 package controlplanemachineset
 
 import (
+	"fmt"
+
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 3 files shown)


pkg/controllers/controlplanemachineset/watch_filters.go
```diff
@@ -71,7 +76,13 @@ func filterControlPlaneMachineSet(namespace string) predicate.Predicate {
 			panic("expected to get an of object of type machinev1.ControlPlaneMachineSet")
 		}
 
-		return cpms.GetNamespace() == namespace && cpms.GetName() == clusterControlPlaneMachineSetName
+		shouldReconcile := cpms.GetNamespace() == namespace && cpms.GetName() == clusterControlPlaneMachineSetName
+
+		if shouldReconcile {
+			klog.V(4).Infof("reconcile triggered be control plane machine set %s/%s", obj.GetNamespace(), obj.GetName())
```
damdo, line 82:
```suggestion
			klog.V(4).Infof("reconcile triggered by control plane machine set %s/%s", obj.GetNamespace(), obj.GetName())
```

damdo (commented):
/lgtm

PR #115, opened 2022-10-12
Add the ability to disable webhook registration

JoelSpeed:
Webhooks cannot be used when debugging locally, every time I'm trying to debug something I have to manually disable the webhook registrationi by commenting out the code.

Then when I build an image to push to test on the cluster, I forget to uncomment.

This change addresses that irritation by allowing me to configure my debug environment to disable the webhooks, without making code changes every time I'm debugging something.

It has no effect on the product but should make debugging easier for us.


damdo (commented):
/lgtm

PR #116, opened 2022-10-14
OCPBUGS-2446: Expose an error condition if an error is repeated more than 15 times in a row

JoelSpeed:
In an effort to improve the observability of errors, this PR sets an error condition on the CPMS status when the reconciler returns an identical error 15 times or more in a row. This happens after approximately 80s since the error first occurs.

This should help reveal spurious errors such as those like when we cannot create a machine because the validation webhooks are failing.

I've manually tested this and it seems to be working as expected.

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -91,6 +91,21 @@ type ControlPlaneMachineSetReconciler struct {
 
 	// ReleaseVersion is the version of current cluster operator release.
 	ReleaseVersion string
+
+	// lastError allows us to track the last error that occurred during reconciliation.
+	lastError *lastErrorTracker
+}
+
+// lastErrorTracker tracks the last error that occurred during reconciliation.
+type lastErrorTracker struct {
+	// lastError is the last error that occurred during reconciliation.
+	lastError error
+
+	// lastErrorTime is the time at which the last error occurred.
+	lastErrorTime metav1.Time
+
+	// count is the number of times we've observed the same error in a row.
+	count int
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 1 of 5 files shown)


pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -320,6 +338,38 @@ func (r *ControlPlaneMachineSetReconciler) reconcileDelete(ctx context.Context,
...
+	case err == nil:
+		// When no error occurred, stop tracking the last error.
+		r.lastError = nil
+	case r.lastError != nil && err.Error() == r.lastError.lastError.Error():
+		// When the error is the same as the last error, increment the count.
+		r.lastError.count++
+	default:
+		// Either the last error is nil or the error is different from the last error.
+		// Track the new error and reset the count.
+		r.lastError = &lastErrorTracker{
```
lobziik, line 354: Will it be better to reuse existing structure, instead of creating new one here?

lobziik (commented):
Overall lgtm, only one nit about structure reusing.

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -160,6 +175,9 @@ func (r *ControlPlaneMachineSetReconciler) Reconcile(ctx context.Context, req ct
 		errs = append(errs, fmt.Errorf("error reconciling control plane machine set: %w", err))
 	}
 
+	// Track the reconcile error as our last error
```
damdo, line 178:
```suggestion
	// Track the reconcile error as our last error.
```

damdo (commented):
LGTM

pkg/controllers/controlplanemachineset/controller.go
lobziik: >What's the concern over initialising a new one?

Nothing aside of mem consumption.
One thing came to my mind, what happen if two different errors will come to this function one after another many times? But, i don't think that this will shoot somewhen in the real usage.

JoelSpeed: > Nothing aside of mem consumption.

Once you overwrite the reference the garbage collector should remove the old instance anyway. I'd rather be safe here than worry about the garbage collection with Go should be doing on our behalf

> One thing came to my mind, what happen if two different errors will come to this function one after another many times? But, i don't think that this will shoot somewhen in the real usage. I'm ok to left it as it is now.

In theory, our controllers are designed to be deterministic and reentrant. So, if they hit one error, so long as it is a persistent error and not transient, we should always go through the same code path over and over and therefore get the same error back. Where we might see issues is when we include variable data such as machine names in error messages, for example the error I fixed here had the name of the machine we created and so each time the error changed. That's something we can control and fix over time if we see bug reports where the error condition didn't catch the issue

pkg/controllers/controlplanemachineset/controller.go
damdo: Was thinking the same, not sure 🤔

PR #117, opened 2022-10-14
Update log messages to use structured logs

JoelSpeed:
These logs messages were recently merged and weren't using structured logs, they should be

### DIFF
--- a/pkg/controllers/controlplanemachineset/watch_filters.go
+++ b/pkg/controllers/controlplanemachineset/watch_filters.go
@@ -46,7 +46,12 @@ const (
 // in the namespace provided.
 func objToControlPlaneMachineSet(namespace string) func(client.Object) []reconcile.Request {
 	return func(obj client.Object) []reconcile.Request {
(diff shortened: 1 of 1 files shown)


pkg/controllers/controlplanemachineset/watch_filters.go
```diff
@@ -79,7 +84,11 @@ func filterControlPlaneMachineSet(namespace string) predicate.Predicate {
 		shouldReconcile := cpms.GetNamespace() == namespace && cpms.GetName() == clusterControlPlaneMachineSetName
 
 		if shouldReconcile {
-			klog.V(4).Infof("reconcile triggered by control plane machine set %s/%s", obj.GetNamespace(), obj.GetName())
+			klog.V(4).Infof(
```
RadekManak, line 87: Should this be Info not Infof?

JoelSpeed: Yes it should, good catch

JoelSpeed:

The ovn and serial jobs both passed, and this is a logging change. I can see the CPMS running in the upgrade test history so I'm content that this PR isn't causing an issue

PR #118, opened 2022-10-14
OCPBUGS-2340: Ensure OnDelete strategy can handle non-standard indexes

JoelSpeed:
There were two issues at hand here.

The first of the two is that the OnDelete strategy was using the list index rather than machine indexes to create new machines. This would mean, if the machine index was 4 but that was the second item in the list, it would ask to create a new Machine in index 1 (indexed from 0). So the second commit here changes the behaviour of the sorted machine info so that it returns a struct with the machine index rather than just a list of the machine infos, this allows us to get the machine index and loop over that instead.

The second issue was in the way we handled indexes that only contain deleting machines within the provider mapping. We were ignoring them at the wrong stage. By ignoring them in the machine mapping generation, there was no opportunity for the output mapping to contain the correct index if that machine had a non-standard index.
Instead, I've moved the ignore to when we are matching candidates, with this approach the deleting machine is a lower priority than non-deleting machines and so when there is a rebalance action, the deleting machine is the one that changes  failure domain rather than a non-deleting machine.


damdo (commented):
Nice 👍 
/lgtm

JoelSpeed:
/test all

/approve

Have been working with @huali9 and have verified that this is working as expected and resolves both OCPBUGS-2340 and OCPBUGS-2349

JoelSpeed:
/retest-required

Installer fix has gone in now

PR #124, opened 2022-10-25
OCPBUGS-2741: Deduplicate failure domains when generating a ControlPlaneMachineSet

JoelSpeed:
On upgrade, when we generate a CPMS, there may be more than the initial set of failure domains in the cluster. To prevent duplication, I've updated the failure domain extraction so that it uses a set symantic and returns each failure domain only once.
This prevents scenarios where the CPMS could be configured with the set `{a, b, a, c}` which we have seen in upgrade tests.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/set.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/set.go
@@ -1,0 +1,70 @@
+/*
+Copyright 2022 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package failuredomain
+
+import "sort"
+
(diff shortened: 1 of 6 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/set.go
```diff
@@ -0,0 +1,70 @@
+/*
```
lobziik, line 1: WDYTH about using a library for this, https://github.com/deckarep/golang-set for example?

lobziik (commented):
almost lgtm, just have a question about using some 3rd party generic set implementation instead of having own one.

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/set.go
JoelSpeed:
```
mySet := mapset.NewSet[T]() // where T is some concrete comparable type.
```
The issue here is that we are creating a set for failure domains, which are an interface and not a concrete or comparable type. This is why we have to use a list and then the `Equal` function rather than using a map.

The set code here is pretty small so I'm not too concerned about maintenance. Perhaps in the future we can change it so that underneath it uses the struct implementation but I think that removes the flexibility of the interface

damdo (commented):
Generally LGTM. Something is wrong with the WH units.

JoelSpeed:
Fixed the webhook issue, the webhooks were expecting to not get a deduplicated list. I think in more places we want deduplicated so I've manually constructed the list for the webhooks

damdo (commented):
/lgtm

JoelSpeed:
QE have verified pre merge
/approve

damdo (commented):
/lgtm

PR #125, opened 2022-10-26
OCPBUGS-2854: Ensure generated CPMS includes any failure domain already present in control plane machines

JoelSpeed:
If the cluster has control plane machines that do not have a failure domain also represented by the machineset failure domains, then the generator does not include the failure domain for the control plane machine and the CPMS cannot be created.

This updates the failure domain gathering logic to include both the Machine and MachineSet failure domains, so that we create a superset from the two.

This should ensure we always include all available failure domains.

### DIFF
--- a/pkg/controllers/controlplanemachinesetgenerator/aws.go
+++ b/pkg/controllers/controlplanemachinesetgenerator/aws.go
@@ -27,6 +27,7 @@ import (
 	machinev1beta1builder "github.com/openshift/client-go/machine/applyconfigurations/machine/v1beta1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/controlplanemachinesetgenerator/aws.go
```diff
@@ -57,14 +58,24 @@ func generateControlPlaneMachineSetAWSSpec(machines []machinev1beta1.Machine, ma
 }
 
 // buildAWSFailureDomains builds an AWSFailureDomain config for the ControlPlaneMachineSet from cluster's MachineSets.
```
damdo, line 60: Machines and MachineSets.

damdo (commented):
/lgtm

PR #129, opened 2022-11-07
[OCPCLOUD-1747] Add RollingUpdate presubmit to test replacement of outdated machine

JoelSpeed:
This PR extends the rolling update replacement periodic test to replace a single instance. This test deliberately only tests a single control plane machine replacement as this should be much quicker than a full rolling update. This is cheaper to run as a presubmit vs the full update strategy that exists today.

This is mostly code motion and a copy of the structure of the periodic, just slimming it down to a 30 minute runtime and replacing a single Machine instead of 3.

### DIFF
--- a/test/e2e/common/common.go
+++ b/test/e2e/common/common.go
@@ -17,6 +17,7 @@ limitations under the License.
 package common
 
 import (
+	"context"
 	"fmt"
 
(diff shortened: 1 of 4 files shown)


damdo (commented):
/lgtm

PR #131, opened 2022-11-09
Add support for Azure e2e tests

JoelSpeed:
Currently, the tests are skipped/fail on Azure because we haven't implemented a way for Azure clusters to be activated (as they are manual) and we haven't added a way to increase the VM size either. This PR does that so that we can E2E test on Azure as well.


damdo (commented):
Nice
/lgtm

JoelSpeed:
There was a small bug in the check for an existing CPMS, I was checking for a not found when I wanted the opposite of that 🤦 Let's try the tests again

PR #137, opened 2022-11-22
[OCPCLOUD-1748] Check OnDelete strategy updates status

JoelSpeed:
This adds a check to the E2E suite that confirms that, when a control plane machine does not have the updated status, the control plane machine set reports the correct status but does not delete the Machine.


damdo (commented):
Nice!
/lgtm

PR #138, opened 2022-11-22
OCPCLOUD-1741: e2e presubmit tests: ControlPlaneMachineSet uninstall

damdo:
This adds a check to the E2E suite that confirms that, when the control plane machine set is deleted, it gets correctly uninstalled and the cluster is cleaned up as expected.

### DIFF
--- a/test/e2e/common/common.go
+++ b/test/e2e/common/common.go
@@ -18,6 +18,7 @@ package common
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -30,6 +31,10 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var (
+	errUIDNotChanged = errors.New("UID has not changed")
 )
 
 // ItShouldHaveAnActiveControlPlaneMachineSet returns an It that checks
@@ -50,6 +55,36 @@ func ExpectControlPlaneMachineSetToBeActive(testFramework framework.Framework) {
 	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
 
 	Expect(cpms.Spec.State).To(Equal(machinev1.ControlPlaneMachineSetStateActive), "control plane machine set should be active")
+}
+
+// ExpectControlPlaneMachineSetToBeInactive gets the control plane machine set and
+// checks that it is active.
+func ExpectControlPlaneMachineSetToBeInactive(testFramework framework.Framework) {
+	By("Checking the control plane machine set is inactive")
+
+	Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
+	k8sClient := testFramework.GetClient()
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+	Expect(cpms.Spec.State).To(Equal(machinev1.ControlPlaneMachineSetStateInactive), "control plane machine set should be inactive")
+}
+
+// ExpectControlPlaneMachineSetToBeInactiveOrNotFound gets the control plane machine set and
+// checks that it is inactive or not found.
+func ExpectControlPlaneMachineSetToBeInactiveOrNotFound(testFramework framework.Framework) {
+	By("Checking the control plane machine set is inactive or not found")
+
+	Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
+	k8sClient := testFramework.GetClient()
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	if err := k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms); err != nil {
+		Expect(err).To(MatchError(ContainSubstring("not found")), "getting control plane machine set should not error")
+	} else {
+		Expect(cpms).To(HaveField("Spec.State", machinev1.ControlPlaneMachineSetStateInactive), "control plane machine set should be inactive or should not exist")
+	}
 }
 
 // EventuallyClusterOperatorsShouldStabilise checks that the cluster operators stabilise over time.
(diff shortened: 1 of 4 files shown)


test/e2e/common/common.go
```diff
@@ -52,6 +57,34 @@ func ExpectControlPlaneMachineSetToBeActive(testFramework framework.Framework) {
...
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+	Expect(cpms.Spec.State).To(Equal(machinev1.ControlPlaneMachineSetStateInactive), "control plane machine set should be inactive")
+}
+
+// ExpectControlPlaneMachineSetToBeInactiveOrNotFound gets the control plane machine set and
+// checks that it is inactive or not found.
+func ExpectControlPlaneMachineSetToBeInactiveOrNotFound(testFramework framework.Framework) {
```
JoelSpeed, line 76: Do you actually use the testFramework here? Looking at other fetch usages, I think using the client may be more used than komega for a Get, so perhaps we should use the framework instead?

test/e2e/common/common.go
```diff
@@ -168,3 +201,79 @@ func WaitForControlPlaneMachineSetDesiredReplicas(ctx context.Context, cpms *mac
...
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).
+		To(Succeed(), "control plane machine set should exist")
+
+	DeleteControlPlaneMachineSet(testFramework, ctx, cpms.DeepCopy())
+
+	if ok := WaitForControlPlaneMachineSetRemovedorRecreated(ctx, cpms); !ok {
```
JoelSpeed, line 238:
```suggestion
	if ok := WaitForControlPlaneMachineSetRemovedOrRecreated(ctx, cpms); !ok {
```

test/e2e/common/common.go
```diff
@@ -168,3 +201,79 @@ func WaitForControlPlaneMachineSetDesiredReplicas(ctx context.Context, cpms *mac
...
+// DeleteControlPlaneMachineSet deletes the control plane machine set.
+func DeleteControlPlaneMachineSet(testFramework framework.Framework, ctx context.Context, cpms *machinev1.ControlPlaneMachineSet) {
+	By("Deleting the control plane machine set")
+
+	k8sClient := testFramework.GetClient()
+	Expect(k8sClient.Delete(ctx, cpms)).To(Succeed(), "control plane machine set should have been deleted")
+}
+
+// WaitForControlPlaneMachineSetRemovedorRecreated waits for the control plane machine set to be removed.
+func WaitForControlPlaneMachineSetRemovedorRecreated(ctx context.Context, cpms *machinev1.ControlPlaneMachineSet) bool {
```
JoelSpeed, line 252:
```suggestion
// WaitForControlPlaneMachineSetRemovedOrRecreated waits for the control plane machine set to be removed.
func WaitForControlPlaneMachineSetRemovedOrRecreated(ctx context.Context, cpms *machinev1.ControlPlaneMachineSet) bool {
```

test/e2e/common/common.go
```diff
@@ -168,3 +201,79 @@ func WaitForControlPlaneMachineSetDesiredReplicas(ctx context.Context, cpms *mac
...
+	By("Waiting for the deleted control plane machine set to be removed/recreated")
+
+	oldUID := cpms.ObjectMeta.UID
+	newCPMS := framework.NewEmptyControlPlaneMachineSet()
+
+	if ok := Eventually(func() error {
+		err := komega.Get(newCPMS)()
```
JoelSpeed, line 259: Use the framework client here I think

test/e2e/common/machine.go
```diff
@@ -256,3 +257,82 @@ func waitForNewMachineRunning(ctx context.Context, oldMachine, newMachine *machi
...
+	for _, machine := range machineList.Items {
+		phase := pointer.StringDeref(machine.Status.Phase, "")
+		if phase != "Running" {
+			notRunningMachines += fmt.Sprintf(",%s", machine.Name)
+		}
+	}
+
+	if notRunningMachines != "" {
+		Fail(fmt.Sprintf("expected all control plane machines in phase running, the following were not: %s", notRunningMachines))
+	}
```
JoelSpeed, line 285: Do you want this to be async? Even if not, you can use some gomega matchers to check each list items meets a condition. The first element to fail will be printed, or, if you don't like the output, you can temporarily register a custom formatter to extract the names

```
Eventually(komega.ObjectList(machineList, machineSelector)).Should(HaveField("Items",
  HaveEach(HaveField("Status.Phase", HaveValue(Equal("Running")))),
))
```

test/e2e/common/machine.go
```diff
@@ -256,3 +257,82 @@ func waitForNewMachineRunning(ctx context.Context, oldMachine, newMachine *machi
...
+func ExpectControlPlaneMachinesNotOwned(testFramework framework.Framework) {
+	By("Checking that none of the control plane machines have owner references")
+
+	machineSelector := runtimeclient.MatchingLabels(framework.ControlPlaneMachineSetSelectorLabels())
+
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+
+	machineList := &machinev1beta1.MachineList{}
+	Expect(k8sClient.List(ctx, machineList, machineSelector)).To(Succeed(), "should be able to list machines")
```
JoelSpeed, line 299: As above, you can do this and the below using Gomega

```
HaveField("Items", HaveEach(
  HaveField("ObjectMeta.OwnerReferences", HaveLen(0)),
))
```

test/e2e/common/machine.go
```diff
@@ -256,3 +257,82 @@ func waitForNewMachineRunning(ctx context.Context, oldMachine, newMachine *machi
...
+func ExpectControlPlaneMachinesWithoutDeletionTimestamp(testFramework framework.Framework) {
+	By("Checking that none of the control plane machines have a deletion timestamp")
+
+	machineSelector := runtimeclient.MatchingLabels(framework.ControlPlaneMachineSetSelectorLabels())
+
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+
+	machineList := &machinev1beta1.MachineList{}
+	Expect(k8sClient.List(ctx, machineList, machineSelector)).To(Succeed(), "should be able to list machines")
```
JoelSpeed, line 325: Use a HaveEach rather than looping yourself

test/e2e/presubmit/presubmit.go
```diff
@@ -117,3 +117,16 @@ func ItShouldRollingUpdateReplaceTheOutdatedMachine(testFramework framework.Fram
...
+
+// ItShouldUninstallTheControlPlaneMachineSet checks that the control plane machine set is correctly uninstalled
+// when a deletion is triggered, without triggering control plane machines changes.
+func ItShouldUninstallTheControlPlaneMachineSet(testFramework framework.Framework) {
+	It("should uninstall the control plane machine set without control plane machine changes", func() {
+		common.ExpectControlPlaneMachineSetToBeInactiveOrNotFound(testFramework)
+		common.ExpectControlPlaneMachinesAllRunning(testFramework)
+		common.ExpectControlPlaneMachinesNotOwned(testFramework)
+		common.ExpectControlPlaneMachinesWithoutDeletionTimestamp(testFramework)
+		common.EventuallyClusterOperatorsShouldStabilise()
```
JoelSpeed, line 129: Doesn't this take some arguments about how long the operators are allowed for stabilisation? Do you need more than the default (1m) or is that ok here?

JoelSpeed:
Your PR title and commit message say `periodic` when they should be `presubmit` 😉

test/e2e/common/machine.go
damdo: I like your suggestion. Thanks.

test/e2e/presubmit/presubmit.go
damdo: I did this on purpose because they should be all stable from the get going (we check this at the beginning with `		common.EventuallyClusterOperatorsShouldStabilise(10*time.Minute, 10*time.Second)`).
And since the uninstall shouldn't cause any destabilisation in the operators (we shouldn't be touching any machine), I expect this to be always stable at this point. So I'm using the common helper with the default 1m, as it should suffice.

test/e2e/common/common.go
```diff
@@ -197,3 +232,81 @@ func EnsureControlPlaneMachineSetUpdateStrategy(testFramework framework.Framewor
...
+	Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
+
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).
+		To(Succeed(), "control plane machine set should exist")
+
+	DeleteControlPlaneMachineSet(testFramework, ctx, cpms.DeepCopy())
```
JoelSpeed, line 267: Nit, didn't expect to be using a DeepCopy here, probably doesn't matter we just don't use them often as they're normally not needed within test suites

test/e2e/common/common.go
```diff
@@ -197,3 +232,81 @@ func EnsureControlPlaneMachineSetUpdateStrategy(testFramework framework.Framewor
...
+
+	if ok := WaitForControlPlaneMachineSetRemovedOrRecreated(ctx, cpms, testFramework); !ok {
+		Fail("timed out waiting for control plane machine set to be removed/recreated")
```
JoelSpeed, line 270: The assertions within `WaitForControlPlaneMachineSetRemovedOrRecreated` will already cause the test to fail, you don'y need to check the result of this function in this case

test/e2e/common/common.go
```diff
@@ -197,3 +232,81 @@ func EnsureControlPlaneMachineSetUpdateStrategy(testFramework framework.Framewor
...
+
+	Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
+
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+	if ok := WaitForControlPlaneMachineSetDesiredReplicas(ctx, cpms); !ok {
```
JoelSpeed, line 248: The assertions within `WaitForControlPlaneMachineSetRemovedOrRecreated` will already cause the test to fail, you don'y need to check the result of this function in this case

test/e2e/common/machine.go
```diff
@@ -256,3 +256,45 @@ func waitForNewMachineRunning(ctx context.Context, oldMachine, newMachine *machi
...
+// have owner references.
+func ExpectControlPlaneMachinesNotOwned(testFramework framework.Framework) {
+	By("Checking that none of the control plane machines have owner references")
+
+	machineSelector := runtimeclient.MatchingLabels(framework.ControlPlaneMachineSetSelectorLabels())
+
+	machineList := &machinev1beta1.MachineList{}
+
+	Eventually(komega.ObjectList(machineList, machineSelector)).Should(HaveField("Items",
+		HaveEach(HaveField("ObjectMeta.OwnerReferences", BeNil())),
```
JoelSpeed, line 284: `BeNil()`? Is that true, what if it were a non-nil list of zero length? Would this still pass? 

Playground suggests we should probably check the length, https://go.dev/play/p/SgoUheVJD1I, it fails only on the last check

damdo: I agree! I changed it to test something and forgot to move it back. Thanks for catching.

test/e2e/common/common.go
damdo: Removed

JoelSpeed (commented):
/approve
/lgtm

PR #140, opened 2022-11-23
[OCPCLOUD-1751] Add integration test for fulling rolling update replacement

JoelSpeed:
This PR makes a number of prefactors:
- Fixing the logging of the CPMS controller suite to use klog
- Allow the testframework to be configured with a namespace (useful for integration tests)
- Allow the rollingupdate tests to be configured via an options struct (so we can set things up in BeforeEach blocks)
- Add a machine manager that pretends to be the machine controller

Then it combines all of these prefactors and the existing periodic test for rolling update replacement to make an integration test that runs through the same logic, but this time, with a fake machine controller instead of a real cluster.


pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -975,6 +981,62 @@ var _ = Describe("With a running controller", func() {
...
+		})
+
+		Context("and the instance size is changed", func() {
+			var testOptions periodic.TestOptions
+
+			BeforeEach(func() {
+				// The CPMS is configured for AWS so use the AWS Platform Type.
+				testFramework := framework.NewFrameworkWith(testScheme, k8sClient, configv1.AWSPlatformType, framework.Full, namespaceName)
+
+				periodic.IncreaseControlPlaneMachineSetInstanceSize(testFramework, 10*time.Second, 1*time.Second)
```
lobziik, line 1028: nit: `periodic` package name is a bit misleading here. Can we do smth about it?

lobziik (commented):
Makes sense to me code wise, the only thing which I don't like - `periodic` module name. If I understand correctly, it's not about periodic tests anymore.

pkg/controllers/controlplanemachineset/controller_test.go
JoelSpeed: Yeah I get what you mean but can't decide how to handle it, I can move the `IncreaseControlPlaneMachineSetInstanceSize` to the common package, but I don't really want to move the actual periodic test from below, WDYT?

Perhaps it's not worth making the distinction between common/periodic/presubmit and just have them in a single folder

CC @damdo WDYT?

damdo: I want to review this Pr. But to your question @JoelSpeed, yes, I think we could be better merging those common/periodic/presubmit helpers, is not easy to forecast if you are going to be using that in both or not. I think having one package is easier.

lobziik: Having it in one package sounds ok to me. I have a feeling that this distinction causes more confusion than profit.

>I think we would be better off merging those common/periodic/presubmit helpers, is not easy to forecast if you are going to be using that in both or not

+1

JoelSpeed: I've moved everything and renamed the package to `helpers`. In doing this I've also moved all of the test cases (`It`) into a single file and grouped everything else by the objects they operate on (CPMS, Machines, ClusterOperators). No code changes to note it's pretty much code motion

lobziik (commented):
/lgtm

lobziik:
/lgtm cancel

I didn't notice unit tests failure, sry

JoelSpeed:
/test unit

There was a flake in the CPMS generator tests, this PR doesn't even touch those 🤔

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -29,6 +30,9 @@ import (
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test"
 	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test/resourcebuilder"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/test/e2e/framework"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/test/e2e/helpers"
```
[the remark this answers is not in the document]
JoelSpeed: Yeah it's not ideal, but the alternative is to rewrite it, right? 

Like, in theory, we should be able to run the whole E2E suite against a fake machine controller, perhaps instead of importing it here, we have a separate integration suite that we run that runs all of the E2E tests against a fake machine controller, WDYT?

I don't really want to rewrite the whole test because we are importing a test from the E2E

JoelSpeed:
Had to rebase this because of conflicts (moving everything to the `helpers` package is a nightmare), but have addressed all of the feedback so far

damdo (commented):
/lgtm

damdo (commented):
/lgtm

PR #141, opened 2022-11-23
OCPCLOUD-1742: e2e presubmit tests: generated ControlPlaneMachineSet replicas updated

damdo:
This adds a check to the E2E suite that confirms that, when the control plane machine set is recreated after its deletion, all the replicas (control plane machines) are up to date as expected.

This PR is based on https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/138 only the last commit is relevant to this one.

### DIFF
--- a/test/e2e/presubmit/presubmit.go
+++ b/test/e2e/presubmit/presubmit.go
@@ -186,3 +186,27 @@ func ItShouldUninstallTheControlPlaneMachineSet(testFramework framework.Framewor
 		common.EventuallyClusterOperatorsShouldStabilise()
 	})
 }
+
+// ItShouldHaveTheControlPlaneMachineSetReplicasUpdated checks that the control plane machine set replicas are updated.
+func ItShouldHaveTheControlPlaneMachineSetReplicasUpdated(testFramework framework.Framework) {
+	It("should have the control plane machine set replicas up to date", func() {
+		By("Checking the control plane machine set replicas are up to date")
+
+		Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
+		k8sClient := testFramework.GetClient()
+
+		cpms := &machinev1.ControlPlaneMachineSet{}
+		Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+		Expect(cpms.Spec.Replicas).ToNot(BeNil(), "replicas should always be set")
+
+		desiredReplicas := *cpms.Spec.Replicas
+
+		Expect(cpms).To(SatisfyAll(
+			HaveField("Status.Replicas", Equal(desiredReplicas)),
+			HaveField("Status.UpdatedReplicas", Equal(desiredReplicas)),
+			HaveField("Status.ReadyReplicas", Equal(desiredReplicas)),
+			HaveField("Status.UnavailableReplicas", Equal(int32(0))),
+		), "control plane machine set replicas should be up to date")
+	})
+}
--- a/test/e2e/presubmit_test.go
+++ b/test/e2e/presubmit_test.go
@@ -89,6 +89,7 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.PreSubmit(), func(
 				})
 
 				presubmit.ItShouldUninstallTheControlPlaneMachineSet(testFramework)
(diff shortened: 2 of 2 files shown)


test/e2e/common/common.go
```diff
@@ -306,3 +306,23 @@ func WaitForControlPlaneMachineSetRemovedOrRecreated(ctx context.Context, cpms *
...
+func ExpectControlPlaneMachineSetReplicasUpdated(testFramework framework.Framework) {
+	By("Checking the control plane machine set replicas are up to date")
+
+	Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
+	k8sClient := testFramework.GetClient()
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+	desiredReplicas := *cpms.Spec.Replicas
```
JoelSpeed, line 321: You can't dereference this before you've check it's not nil, you'll need the `Spec.Replace` not nil check before this

test/e2e/common/common.go
```diff
@@ -306,3 +306,23 @@ func WaitForControlPlaneMachineSetRemovedOrRecreated(ctx context.Context, cpms *
...
+
+	Expect(cpms).To(SatisfyAll(
+		HaveField("Spec.Replicas", Not(BeNil())),
+		HaveField("Status.Replicas", Equal(desiredReplicas)),
+		HaveField("Status.UpdatedReplicas", Equal(desiredReplicas)),
```
JoelSpeed, line 326: Do you also want to check `ReadyReplicas` and `UnavailableReplicas` since we expect those to be `desiredReplicas` and 0 respectively

test/e2e/presubmit/presubmit.go
```diff
@@ -186,3 +186,10 @@ func ItShouldUninstallTheControlPlaneMachineSet(testFramework framework.Framewor
 		common.EventuallyClusterOperatorsShouldStabilise()
 	})
 }
+
+// ItShouldHaveTheControlPlaneMachineSetReplicasUpdated checks that the control plane machine set replicas are updated.
+func ItShouldHaveTheControlPlaneMachineSetReplicasUpdated(testFramework framework.Framework) {
+	It("should have the control plane machine set replicas up to date", func() {
+		common.ExpectControlPlaneMachineSetReplicasUpdated(testFramework)
```
JoelSpeed, line 193: I wonder if we should just embed the content of this function directly here? We are doing it in other places where the code isn't overly complex and I'd say this is a relatively simple check

test/e2e/common/common.go
damdo: Oh yes, good spot.

test/e2e/presubmit/presubmit.go
```diff
@@ -186,3 +186,28 @@ func ItShouldUninstallTheControlPlaneMachineSet(testFramework framework.Framewor
...
+
+		cpms := &machinev1.ControlPlaneMachineSet{}
+		Expect(k8sClient.Get(testFramework.GetContext(), framework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+		Expect(cpms.Spec.Replicas).ToNot(BeNil(), "replicas should always be set")
+
+		desiredReplicas := *cpms.Spec.Replicas
+
+		Expect(cpms).To(SatisfyAll(
+			HaveField("Spec.Replicas", Not(BeNil())),
```
JoelSpeed, line 206: Do we need to check this again? It's now done on 201

test/e2e/presubmit/presubmit.go
```diff
@@ -186,3 +186,28 @@ func ItShouldUninstallTheControlPlaneMachineSet(testFramework framework.Framewor
...
+			HaveField("Status.Replicas", Equal(desiredReplicas)),
+			HaveField("Status.UpdatedReplicas", Equal(desiredReplicas)),
+			HaveField("Status.ReadyReplicas", Equal(desiredReplicas)),
+			HaveField("Status.UnavailableReplicas", Equal(0)),
```
JoelSpeed, line 210: I'd expect this to type mismatch, you probably need to cast the 0 as an `int32`
```suggestion
			HaveField("Status.UnavailableReplicas", Equal(int32(0))),
```

PR #142, opened 2022-11-24
Ensure stable output from RunCheckUntil

JoelSpeed:
This test has been flaking because there was an unstable output in the case the context was cancelled before the timeout was reached.

Either the `Eventually` would catch the timeout in which case the test passed, or, the `checkErr` would be `errContextCancelled` in which case the `StopTrying` returned and the test failed.

By intercepting `errContextCancelled` and returning it plainly, the `Eventually` is always the one catching the context timeout, meaning we now have stable errors.

I also added some context to the `Eventually` so that when it fails it includes the message `check failed or condition did not succeed before the context was cancelled`


damdo (commented):
/lgtm

PR #145, opened 2022-11-25
OCPCLOUD-1750: e2e presubmit test: activating ControlPlaneMachineSet adds owner references

damdo:
This adds a check to the E2E suite that confirms that, when the machines are up to date and the control plane machine set is activated, the correct owner references are added.


damdo:
/hold

until https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/140 merges

damdo:
/unhold
#140 merged

PR #146, opened 2022-11-29
[OCPCLOUD-1754] Check OnDelete strategy rebalances machines

JoelSpeed:
This PR adds a new integration test that checks that the OnDelete strategy correctly balances machines across availability zones.

It starts with all Machines in the same zone, and deletes a machine. For cases where the 0, or 1 index is deleted first, this means the machine goes into zone B, and the next deleted machine goes into zone C.

When index 2 is deleted first, it will always go into zone C as the standard mapping is used (`[0: a, 1: b, 2:c]`). The next deleted machine will go into zone B

### DIFF
--- a/pkg/controllers/controlplanemachineset/controller_test.go
+++ b/pkg/controllers/controlplanemachineset/controller_test.go
@@ -19,6 +19,7 @@ package controlplanemachineset
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -36,6 +37,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


Fedosin (approved):
/lgtm

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -1044,6 +1046,125 @@ var _ = Describe("With a running controller", func() {
...
+				Build()
+			Expect(k8sClient.Create(ctx, cpms)).Should(Succeed())
+
+			By("Waiting for the ControlPlaneMachineSet to report a stable status")
+			Eventually(komega.Object(cpms)).Should(HaveField("Status", SatisfyAll(
+				HaveField("Replicas", Equal(int32(3))),
+				HaveField("UpdatedReplicas", Equal(int32(1))),
+				HaveField("ReadyReplicas", Equal(int32(3))),
+				HaveField("UnavailableReplicas", Equal(int32(0))),
+			)))
```
damdo, line 1094: Maybe we could argument with a comment what do we mean by stable status here and why the update count is 1.

damdo (commented):
LGTM, just one suggestion.

pkg/controllers/controlplanemachineset/controller_test.go
JoelSpeed: Good suggestion, added a comment

damdo (commented):
Thanks!
/lgtm

PR #147, opened 2022-11-30
OCPCLOUD-1744: e2e presubmit: update to newest machine triggers inactive ControlPlaneMachineSet regeneration

damdo:
This adds a check to the E2E suite that confirms that, when the reference machine (newest, alphabetically latest) provider spec is updated, and the `ControlPlaneMachineSet` is `Inactive`, the `ControlPlaneMachineSet` is regenerated by the `control-plane-machine-set-generator` controller.

### DIFF
--- a/test/e2e/helpers/cases.go
+++ b/test/e2e/helpers/cases.go
@@ -39,9 +39,8 @@ func ItShouldHaveAnActiveControlPlaneMachineSet(testFramework framework.Framewor
 	})
 }
 
-// TestOptions allow the test cases to be configured with a test framework and
-// optional timeouts.
-type TestOptions struct {
+// RollingUpdatePeriodicTestOptions allow the test cases to be configured.
+type RollingUpdatePeriodicTestOptions struct {
 	TestFramework        framework.Framework
 	RolloutTimeout       time.Duration
 	StabilisationTimeout time.Duration
@@ -49,7 +48,7 @@ type TestOptions struct {
 
 // ItShouldPerformARollingUpdate checks that the control plane machine set performs a rolling update
 // in the manner desired.
-func ItShouldPerformARollingUpdate(opts *TestOptions) {
+func ItShouldPerformARollingUpdate(opts *RollingUpdatePeriodicTestOptions) {
 	It("should perform a rolling update", Offset(1), func() {
 		Expect(opts).ToNot(BeNil(), "test options are required")
 		Expect(opts.TestFramework).ToNot(BeNil(), "testFramework is required")
--- a/test/e2e/helpers/cases.go
+++ b/test/e2e/helpers/cases.go
@@ -24,6 +24,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
 
 	machinev1 "github.com/openshift/api/machine/v1"
@@ -44,6 +45,15 @@ type RollingUpdatePeriodicTestOptions struct {
 	TestFramework        framework.Framework
 	RolloutTimeout       time.Duration
 	StabilisationTimeout time.Duration
+}
+
+// ControlPlaneMachineSetRegenerationTestOptions allow test cases to be configured.
+type ControlPlaneMachineSetRegenerationTestOptions struct {
+	TestFramework        framework.Framework
+	OriginalProviderSpec machinev1beta1.ProviderSpec
+	UpdatedProviderSpec  machinev1beta1.ProviderSpec
+	UID                  types.UID
+	Index                int
 }
 
 // ItShouldPerformARollingUpdate checks that the control plane machine set performs a rolling update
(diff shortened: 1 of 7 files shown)


test/e2e/helpers/cases.go
```diff
@@ -302,3 +309,43 @@ func ItShouldCheckAllControlPlaneMachinesHaveCorrectOwnerReferences(testFramewor
...
+	It("should perform control plane machine set regeneration", func() {
+		Expect(opts.TestFramework).ToNot(BeNil(), "test framework should not be nil")
+		ctx := opts.TestFramework.GetContext()
+		cpms := opts.TestFramework.NewEmptyControlPlaneMachineSet()
+
+		// Check that the control plane machine set is regenerated.
+		WaitForControlPlaneMachineSetRemovedOrRecreated(ctx, opts.TestFramework, opts.UID)
+		EnsureInactiveControlPlaneMachineSet(opts.TestFramework)
+
+		// Check that the control plane machine set reports 1 updated machine (the other two will need update).
```
JoelSpeed, line 325: Does this comment not just duplicate the `By` below? probably can remove it

test/e2e/helpers/cases.go
```diff
@@ -302,3 +309,43 @@ func ItShouldCheckAllControlPlaneMachinesHaveCorrectOwnerReferences(testFramewor
...
+			machine.Spec.ProviderSpec = opts.ProviderSpec
+		})}, gomegaArgs...)
+		Eventually(updateMachineArgs...).Should(Succeed(), "control plane machine should be able to be updated")
+
+		// Check that the control plane machine set is regenerated.
+		WaitForControlPlaneMachineSetRemovedOrRecreated(ctx, opts.TestFramework, opts.UID)
+		EnsureInactiveControlPlaneMachineSet(opts.TestFramework)
+
+		// Check that all control plane machine set replicas are up to date.
+		EnsureControlPlaneMachineSetUpdated(opts.TestFramework)
```
JoelSpeed, line 349: This to me feels like an `AfterAll` in the ordered block containing this test. The setup of the machine being new is in the `BeforeAll`, so the cleanup should be similar

test/e2e/helpers/cases.go
```diff
@@ -302,3 +309,43 @@ func ItShouldCheckAllControlPlaneMachinesHaveCorrectOwnerReferences(testFramewor
...
+		ctx := opts.TestFramework.GetContext()
+		cpms := opts.TestFramework.NewEmptyControlPlaneMachineSet()
+
+		// Check that the control plane machine set is regenerated.
+		WaitForControlPlaneMachineSetRemovedOrRecreated(ctx, opts.TestFramework, opts.UID)
+		EnsureInactiveControlPlaneMachineSet(opts.TestFramework)
+
+		// Check that the control plane machine set reports 1 updated machine (the other two will need update).
+		By("Checking the control plane machine set reports 1 updated machine, 2 needing update")
+		Eventually(komega.Object(cpms)).Should(
```
JoelSpeed, line 327: Do we want to check that the CPMS gets the expected `providerSpec`?

test/e2e/presubmit_test.go
```diff
@@ -33,6 +33,10 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.PreSubmit(), func(
 		helpers.EventuallyClusterOperatorsShouldStabilise(10*time.Minute, 10*time.Second)
 	}, OncePerOrdered)
 
+	AfterEach(func() {
+		helpers.EnsureActiveControlPlaneMachineSet(testFramework)
+	}, OncePerOrdered)
```
JoelSpeed, line 38: Wondering if we need this? Given each test we've written so far has a Before that sets up the Active/Inactive state of the CPMS 🤔

damdo: I had to add this because in the common test suite we are expecting this and if the presubmit suite is executed first then there is a chance we exit the suite with an inactive cpms, which causes the expectation in the common suite to fail.

JoelSpeed: Hmm, good point, the problem then is that the common test doesn't really work. We really need that to be the first test otherwise other tests in the suite are going to break/interfere with it.

On a manual support cluster, we might end up activating the CPMS before that test starts, which isn't the expected state

Maybe that test isn't really that useful?

test/e2e/helpers/cases.go
damdo: In the `BeforeAll` we have the setup yes, and I'd expect in the `AfterAll` to have cleanup/stabilization.
Although I think making sure the CPMS is updated after a regeneration should be part of the test itself rather than something that happens afterwards in a separate phase.

JoelSpeed: > Although I think making sure the CPMS is updated after a regeneration should be part of the test itself rather than something that happens afterwards in a separate phase.

Yes, I agree, so I'm seeing this as three phases

* Set up
  * Make sure a CPMS exists
  * Make sure the CPMS is updated
  * Update newest machine provider spec
* Test
  * Check eventually new CPMS is created
  * Check new CPMS has expected providerSpec
  * Check new CPMS has expected status
* Tear down
  * Set newest machine provider spec back to expected
  * Wait for CPMS to report updated again

test/e2e/presubmit_test.go
JoelSpeed: So long as on platforms where we expect the generator to create a CPMS, it is doing so, then that's ok with me, and I believe we have checks for that already through the rest of the tests

test/e2e/helpers/cases.go
damdo: Yes, that'd be good.

test/e2e/presubmit_test.go
damdo: Let's do this in a separate PR, once this merged.

test/e2e/helpers/cases.go
damdo: Added.

JoelSpeed (commented):
/lgtm
/approve

@damdo I'm assigning you the cleanup of the common test that you've suggested will be better served in a follow up PR, please make sure that gets resolved

damdo:
The Cluster Operators stabilisation at setup timed out.
/retest-required

damdo:
The [e2e-aws-operator step passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-control-plane-machine-set-operator/147/pull-ci-openshift-cluster-control-plane-machine-set-operator-main-e2e-aws-operator/1598308729097293824#1:build-log.txt%3A47) but the job [failed on cluster deprovisioning](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-control-plane-machine-set-operator/147/pull-ci-openshift-cluster-control-plane-machine-set-operator-main-e2e-aws-operator/1598308729097293824)

Thus overriding.
/override ci/prow/e2e-aws-operator

PR #155, opened 2022-12-22
Add testing strategy documentation

JoelSpeed:
We developed a [testing strategy](https://docs.google.com/document/d/1nBwmLxDQjV8AZE0XDThp8W2meh9ZPTrX6NxazNMRAGU/edit#heading=h.idcx9qyyu014) document while working out what E2Es we wanted, however this document was previously internal only. I think it should be immortalised in git for the world to see (and so that we can version control it as we make changes in the future.


elmiko (commented):
/lgtm

PR #156, opened 2022-12-23
OCPCLOUD-1740: e2e periodic test: machine replacement with cluster wide proxy

damdo:
This adds a periodic check to the E2E suite that confirms that, when a cluster wide proxy is configured, the CPMSO is able to create machine replacements, and control plane machines are correctly provisioned by MAPI.

### DIFF
--- a/test/e2e/helpers/proxy.go
+++ b/test/e2e/helpers/proxy.go
@@ -1,0 +1,391 @@
+/*
+Copyright 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helpers
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apitypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	client "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/pkg/test/resourcebuilder"
+	"github.com/openshift/cluster-control-plane-machine-set-operator/test/e2e/framework"
+)
+
+const (
+	mapiControllersDeploymentName         = "machine-api-controllers"
+	machineControllerContainerName string = "machine-controller"
+	proxyName                             = "mitm-proxy"
+	mitmSignerName                        = "mitm-signer"
+	mitmBootstrapName                     = "mitm-bootstrap"
+	mitmCustomPKIName                     = "mitm-custom-pki"
+	mitmCustomPKINamespace                = "openshift-config"
+	mitmDaemonsetName                     = proxyName
+	mitmServiceName                       = proxyName
+)
+
+const proxySetup = `
+cd /.mitmproxy
+cat /root/certs/tls.key /root/certs/tls.crt > /.mitmproxy/mitmproxy-ca.pem
+curl -O https://snapshots.mitmproxy.org/5.3.0/mitmproxy-5.3.0-linux.tar.gz
+tar xvf mitmproxy-5.3.0-linux.tar.gz
+HOME=/.mitmproxy ./mitmdump --ssl-insecure
+`
+
+// DeployProxy deploys a MITM Proxy to the cluster.
+func DeployProxy(testFramework framework.Framework, gomegaArgs ...interface{}) {
+	proxyNamespace := testFramework.ControlPlaneMachineSetKey().Namespace
+	proxyLabels := map[string]string{
+		"app": proxyName,
+	}
+
+	// Generate an RSA private key and its corresponding X.509 certificate,
+	// for the MITM proxy.
+	certBytes, keyBytes, err := generateCert()
+	Expect(err).NotTo(HaveOccurred())
+
+	mitmSignerKey := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: keyBytes})
+	mitmSignerCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certBytes})
+
+	mitmSignerSecret := resourcebuilder.Secret().WithName(mitmSignerName).WithNamespace(proxyNamespace).WithLabels(proxyLabels).
+		WithData(map[string][]byte{"tls.crt": mitmSignerCert, "tls.key": mitmSignerKey}).Build()
+
+	mitmBootstrapConfigMap := resourcebuilder.ConfigMap().WithName(mitmBootstrapName).WithNamespace(proxyNamespace).WithLabels(proxyLabels).
+		WithData(map[string]string{"startup.sh": proxySetup}).Build()
+
+	mitmCustomPkiConfigMap := resourcebuilder.ConfigMap().WithName(mitmCustomPKIName).WithNamespace(mitmCustomPKINamespace).
+		WithData(map[string]string{"ca-bundle.crt": string(mitmSignerCert)}).Build()
+
+	mitmDaemonset := resourcebuilder.DaemonSet().WithName(proxyName).WithNamespace(proxyNamespace).WithLabels(proxyLabels).
+		WithVolumes(buildDaemonSetVolumes()).WithContainers(buildDaemonSetContainers()).Build()
+
+	mitmService := resourcebuilder.Service().WithNamespace(proxyNamespace).WithName(proxyName).
+		WithLabels(proxyLabels).WithSelector(proxyLabels).WithPorts(buildServicePorts()).Build()
+
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+
+	By("Creating the MITM proxy Secret")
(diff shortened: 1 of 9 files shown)


test/e2e/presubmit_test.go
```diff
@@ -38,6 +38,27 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.PreSubmit(), func(
 			helpers.EnsureActiveControlPlaneMachineSet(testFramework)
 		}, OncePerOrdered)
 
+		Context("with a cluster wide proxy", Ordered, func() {
```
JoelSpeed, line 41: This looks good but can be moved to the periodic file :)

test/e2e/helpers/proxy.go
```diff
@@ -0,0 +1,362 @@
...
+	Eventually(komega.Object(mitmService), time.Minute*1).Should(
+		HaveField("Spec.ClusterIP", Not(Equal(""))),
+	)
+}
+
+// ConfigureClusterWideProxy configures the Cluster-Wide Proxy to use the MITM Proxy.
+func ConfigureClusterWideProxy(testFramework framework.Framework, gomegaArgs ...interface{}) {
+	k8sClient := testFramework.GetClient()
+	ctx := testFramework.GetContext()
+	proxyNamespace := testFramework.ControlPlaneMachineSetKey().Namespace
```
JoelSpeed, line 105: Thinking about the future, and this doesn't need to be done now, but we may want to change the inteface up a bit so that for actuator pkg it has a `GetNamespace()`, then the CPMS specific framework can embed the generic one 🤔

test/e2e/helpers/proxy.go
```diff
@@ -0,0 +1,362 @@
...
+
+	By("Waiting for machine-api-controller deployment to reflect configured cluster-wide proxy")
+
+	Eventually(komega.Object(deploy), time.Minute*5).Should(
+		HaveField("Spec.Template.Spec.Containers", ContainElement(SatisfyAll(
+			HaveField("Name", Equal(machineControllerContainerName)),
+			HaveField("Env", SatisfyAll(
+				ContainElement(SatisfyAll(
+					HaveField("Name", "NO_PROXY"),
+					HaveField("Value", Not(BeNil())),
```
JoelSpeed, line 140: Is `Value` a pointer?

test/e2e/helpers/proxy.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	By("Waiting for machine-api-controller deployment to reflect unconfigured cluster-wide proxy")
+	Eventually(komega.Object(deploy), time.Minute*5).Should(
+		HaveField("Spec.Template.Spec.Containers", ContainElement(SatisfyAll(
+			HaveField("Name", Equal(machineControllerContainerName)),
+			HaveField("Env", SatisfyAll(
+				Not(ContainElement(SatisfyAll(
+					HaveField("Name", "NO_PROXY"),
+					HaveField("Value", BeFalse()),
```
JoelSpeed, line 186: This confuses me, what's the expectation here? Not sure why it's `BeFalse()`? Should the expectation be that it shouldn't contain an element with name `NO_PROXY` and that would be sufficient?

test/e2e/helpers/proxy.go
```diff
@@ -0,0 +1,362 @@
...
++pIrgIxm2p5wQmSGgI5KNUL6dgU3dC5aRDMU7hV0RGXKrJktTmU2SnZsOUe4SZ5L
+Vyanpk8N1QIDAQABo1MwUTAdBgNVHQ4EFgQUFD2D/LygThvwLRKukmH1Y++hxLIw
+HwYDVR0jBBgwFoAUFD2D/LygThvwLRKukmH1Y++hxLIwDwYDVR0TAQH/BAUwAwEB
+/zANBgkqhkiG9w0BAQsFAAOBgQCLZeGgfFRCOWqSXcDzUzeJE03HD/y2rLY+hDMn
+kj69xY3+K0z4DRbb1vpMiYsTYF/Z1d6blND4KW+7oi9R6PPYg6XATDMf9tVfTQIe
+qBkDNfqABPlwJABpixwD20XXQUBqADEyO3tdQLtiMi5Qr6QHOX3+FepiHAgdxAFt
+Mqy4Gw==
+-----END CERTIFICATE-----
+`
+	mitmSignerKey = `
```
JoelSpeed, line 345: How hard would it be to generate a fresh key and cert here rather than embedding something? 

The idea of embedded certs throws up red flags, even when it's in test code.

Something like, https://github.com/oauth2-proxy/oauth2-proxy/blob/f753ec1ca5c42cd4c4e400afa1f76817638f3010/pkg/util/util.go#L35-L70, might be helpful?

test/e2e/helpers/proxy.go
damdo: Leftover, will remove.

test/e2e/helpers/proxy.go
damdo: I copied the approach we have in the other implementation of this, but I agree this would be nicer.
Will look into it. Thanks.

test/e2e/helpers/proxy.go
damdo: This was something I wanted to consult you on. It works but I think there is something better for this, probaby: `Not(BeEmpty())`?

test/e2e/presubmit_test.go
damdo: Yes, that's the plan 👍

test/e2e/helpers/proxy.go
JoelSpeed: Have you tried `HaveValue(Not(BeEmpty))`?

test/e2e/helpers/proxy.go
damdo: Agree we should do this 👍

test/e2e/helpers/proxy.go
damdo: Fixed 👍

test/e2e/helpers/proxy.go
damdo: Removed

test/e2e/helpers/proxy.go
damdo: Changed approach to generate this dynamically.

JoelSpeed:
Changes LGTM, please make sure this is in the periodics and not presubmits and then we can merge

test/e2e/presubmit_test.go
damdo: Moved.

test/e2e/periodic_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 Red Hat, Inc.
+Copyright 2023 Red Hat, Inc.
```
JoelSpeed, line 2: You only need to update new files, old files are checked against their creation timestamp for when this should be

test/e2e/presubmit_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 Red Hat, Inc.
+Copyright 2023 Red Hat, Inc.
```
JoelSpeed, line 2: No need to update old files, the linter only checks the year on new files in a PR

test/e2e/periodic_test.go
damdo: Screenshot 2023-01-03 at 14 19 25
`goheader` was reporting this. That's why I changed it

test/e2e/presubmit_test.go
damdo: Same as: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/156#discussion_r1060578285

test/e2e/periodic_test.go
JoelSpeed: Huh, weird, it's supposed to ignore them on files that haven't been created, is your IDE using the config from the project? Do you get the same result if you run `make lint`?

damdo: Yeah, I do get the same result with `make lint`:
```
[~/src/cluster-control-plane-machine-set-operator git:(e2e-machine-replacement-with-proxy*)] $ make lint
WARN [runner] The linter 'ifshort' is deprecated (since v1.48.0) due to: The repository of the linter has been deprecated by the owner.
WARN [linters context] wastedassign is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649.
test/e2e/periodic_test.go:1:11: Expected:2023, Actual: 2022 Red Hat, Inc. (goheader)
/*
  ^
exit status 1
make: *** [lint] Error 1
```

JoelSpeed: Yeah ok, never mind then, we can update them

JoelSpeed:
The CI failed because operators weren't stable within time, have we looked into why they failed? Is there a genuine issue or is the timeout just not sufficient for uninstalling the proxy?

damdo:
I looked into it, I've always (~ 5/6 times) seen it passing locally. I believe it's a timing issue on CI. So I wanted to see if it was hitting it again. That said I believe we should bump the timeout in this kind of situation.

damdo:
Cluster creation failed.
/test e2e-aws-periodic-pre

test/e2e/periodic_test.go
```diff
@@ -35,6 +35,27 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
 			helpers.EnsureActiveControlPlaneMachineSet(testFramework)
 		})
 
+		Context("with a cluster wide proxy", Ordered, func() {
+			BeforeAll(func() {
+				helpers.DeployProxy(testFramework)
+				helpers.ConfigureClusterWideProxy(testFramework)
+				helpers.EventuallyClusterOperatorsShouldStabilise(1*time.Hour, 10*time.Second)
```
JoelSpeed, line 42: The cluster should probably be stable before the test starts, I wouldn't expect to need to wait this long

JoelSpeed:
> {"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:164","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 2h0m0s timeout","severity":"error","time":"2023-01-10T14:29:52Z"}

Looks like we may need to increase the timeout for the periodic tests, @damdo can you look into making them a 3h limit?

JoelSpeed:
/approve

This looks good in general and I can see the tests are passing, good work, lets try and make sure we can get the limit increased though else you'll break our periodics

damdo:
Bumping periodics timeout to 3h here: https://github.com/openshift/release/pull/35271
cc. @JoelSpeed

damdo:
Network flake, retesting.

/test e2e-aws-periodic-pre
/test e2e-aws-upgrade

JoelSpeed:
Looks like this hit the ginkgo timeout this time, https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/hack/e2e.sh#L12, we will need to up that one too

damdo:
We know this test is broken and have spoken to the etcd team about it, they are happy for us to break them temporarily so that we can fix the defect in the product.

PR #165, opened 2023-02-01
OCPBUGS-6760: webhooks: disable mandatory TargetPools validation on GCP

damdo:
This causes issue for private GCP clusters.
We will re-add this check as a warning in the future, when [webhook warnings
are supported by controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/pull/2014). See: https://issues.redhat.com/browse/OCPCLOUD-1894


JoelSpeed:
/approve
/lgtm

Thanks for being on top of this and making sure we track reverting this later

PR #169, opened 2023-02-14
Update ginkgo to 2.8.1

RadekManak:
The race condition bug in `by` introduced in ginkgo 2.5.0 should be fixed by 2.8.1. We can upgrade now.


damdo (commented):
/approve

Thanks!

PR #170, opened 2023-02-16
e2e: periodics: remove test with cluster-wide-proxy

damdo:
Following discussions on the 9th Feb 2023 during the [Cluster Lifecycle arch call](https://docs.google.com/document/d/1RAkb9QL8Gg7USFiuZi2RpgTZwi-ieTDCdG4YOOkxWaE) we decided that this test is too disruptive for us as the cluster operators recovery from the cluster-wide-proxy setup takes too long and break the overall allotted time for this test. Furthermore we are already testing machine creation under cluster-wide-proxy within the
machine-api E2Es.

As such we are removing this test and will collaborate with the wider OpenShift team to resurface a cluster-wide-proxy test for the entire OpenShift when/if that comes to fruition.


JoelSpeed:
Has the helpers code been migrated over to actuator pkg to replace the old test that is in there?

damdo:
Not sure if @RadekManak ported the proxy helper as well.

JoelSpeed:
Should we in tandem to this, copy the code over and rewrite the actuator pkg version of this test to use the new utils you are removing from here?

RadekManak:
> Not sure if @RadekManak ported the proxy helper as well.

I only moved stuff from pkg/test. The test/e2e/framework has dependency on ProviderConfig. We would need to decouple that if we want to move some of the ginkgo steps over.

PR #171, opened 2023-02-27
OCPBUGS-7989: machine's node must be ready for CPMS machine to be ready

damdo:
A CPMS Machine should be considered Ready only when both its underlying MAPI Machine is `Running` **and** the linked Node is `Ready`.

### DIFF
--- a/test/integration/machine_manager.go
+++ b/test/integration/machine_manager.go
@@ -25,7 +25,9 @@ import (
 
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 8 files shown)


test/integration/machine_manager.go
```diff
@@ -201,6 +212,47 @@ func (r *integrationMachineManager) setPhase(ctx context.Context, logger logr.Lo
...
+					// The node must be Ready for the CPMS replica
+					// to be considered Ready.
+					Type:   corev1.NodeReady,
+					Status: corev1.ConditionTrue,
+				},
+			},
+		},
+	}
+
+	if err := r.Create(ctx, &node); err != nil {
```
JoelSpeed, line 239: Does a node not have a status subresource? Do you not need to first create the node and then update the status on it?

damdo: Yes, the Node has a status subresource, although I noticed that creating a Node with the status already filled in, is also a viable option, as conditions are retained.

damdo:
Azure is expected to fail due to an existing issue independent of this PR: [OCPBUGS-7359](https://issues.redhat.com/browse/OCPBUGS-7359)

PR #175, opened 2023-03-09
OCPBUGS-2960: Check ProviderSpec before generating MachineInfo

odvarkadaniel:
If we remove the `vmSize` field from the CPMS ProviderSpec on Azure, the master machine gets stuck in a loop of provisiong/deleting states.
We try to fix this by adding a defaulting webhook for Azure's ProviderSpec, specifically for the `vmSize` field. If we see that this field is empty, we provide the ProviderSpec with a default value for that field.

- [x] Propose a solution
- [ ] Come up with reasonable pre-submit E2E test covering this
- [x] Test this on a running Azure cluster

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -100,7 +100,7 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
 		return nil, errEmptyConfig
 	}
 
-	providerConfig, err := providerconfig.NewProviderConfigFromMachineTemplate(*cpms.Spec.Template.OpenShiftMachineV1Beta1Machine)
+	providerConfig, err := buildValidProviderConfig(ctx, cl, cpms)
 	if err != nil {
 		return nil, fmt.Errorf("error building a valid provider config: %w", err)
 	}
@@ -134,6 +134,43 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
 		namespace:            cpms.Namespace,
 		machineAPIScheme:     machineAPIScheme,
 	}, nil
+}
+
+// buildValidProviderConfig makes sure that the provider config is valid by dry-run creating a machine.
+func buildValidProviderConfig(ctx context.Context, cl client.Client, cpms *machinev1.ControlPlaneMachineSet) (providerconfig.ProviderConfig, error) {
+	machine := &machinev1beta1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "tmp-machine-dry-run",
+			Namespace:    cpms.Namespace,
+			Annotations:  cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.ObjectMeta.Annotations,
+			Labels:       cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.ObjectMeta.Labels,
+		},
+		Spec: cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.Spec,
+	}
+
+	providerConfig, err := providerconfig.NewProviderConfigFromMachineTemplate(*cpms.Spec.Template.OpenShiftMachineV1Beta1Machine)
+	if err != nil {
+		return nil, fmt.Errorf("could not create a new provider config from machine template: %w", err)
+	}
+
+	rawConfig, err := providerConfig.RawConfig()
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch raw config from provider config: %w", err)
+	}
+
+	machine.Spec.ProviderSpec.Value.Raw = rawConfig
+
+	dryRunClient := client.NewDryRunClient(cl)
+	if err := dryRunClient.Create(ctx, machine); err != nil {
+		return nil, fmt.Errorf("could not dry-run create the machine: %w", err)
+	}
+
+	machineProviderConfig, err := providerconfig.NewProviderConfigFromMachineSpec(machine.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("could not get provider config for machine: %w", err)
+	}
+
+	return machineProviderConfig, nil
 }
 
 // openshiftMachineProvider holds the implementation of the MachineProvider interface.
@@ -201,10 +238,6 @@ func (m *openshiftMachineProvider) GetMachineInfos(ctx context.Context, logger l
 	}
 
 	for _, machine := range machineList.Items {
-		if err := m.ensureValidProviderConfig(ctx); err != nil {
-			return nil, fmt.Errorf("could not ensure the validity of a provider config: %w", err)
-		}
-
 		machineInfo, err := m.generateMachineInfo(ctx, logger, machine)
 		if err != nil {
 			return nil, fmt.Errorf("could not generate machine info for machine %s: %w", machine.Name, err)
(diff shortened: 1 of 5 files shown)


odvarkadaniel:
Work is still in progress. Have to research a little bit more if this is the right way to do this. Also don't know if this is a working solution at all.

/hold

odvarkadaniel:
/retitle OCPBUGS-2960: Check ProviderSpec before generating MachineInfo

odvarkadaniel:
Tested this change on a running Azure cluster and seems to be working properly - the new machine gets to state `Running` with a default `vmSize = Standard_D4s_v3`.

odvarkadaniel:
Just tested this fix on AWS and Azure and both providers work fine. For defaulted values that are the same, no rollout happens. For defaulted values that are different (e.g. vmSize is set to Standard_D8s_v3, but defaulted value is Standard_D4s_v3) a rollout happens for all the master machines.

This PR also introduces an e2e-presubmit test, where we delete a defaulted value by the MAO webhook and we expect to not cause a rollout. This was successfully tested on an Azure cluster.

Thank you for looking into this!
/cc @zhaozhanqi @JoelSpeed @damdo

test/e2e/helpers/controlplanemachineset.go
```diff
@@ -328,3 +328,27 @@ func GetControlPlaneMachineSetUID(testFramework framework.Framework) types.UID {
 
 	return cpms.ObjectMeta.UID
 }
+
+// MakeControlPlaneMachineSetProviderConfigInvalid makes the current ControlPlaneMachineSet's providerSpec invalid
+// by removing a defaulted value.
+func MakeControlPlaneMachineSetProviderConfigInvalid(testFramework framework.Framework, gomegaArgs ...interface{}) machinev1beta1.ProviderSpec {
```
JoelSpeed, line 334: I'd probably rename this, to me, this suggests we want to make it broken so it doesn't work, but, we actually want to test defaulting, so perhaps we can remove the invalid notion and reword this a bit?

test/e2e/presubmit_test.go
```diff
@@ -102,6 +102,19 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.PreSubmit(), func(
 				})
 			})
 		})
+
+		Context("and the defaulted value delete from the ControlPlaneMachineSet", func() {
```
JoelSpeed, line 106:
```suggestion
		Context("and a defaulted value is deleted from the ControlPlaneMachineSet", func() {
```

test/e2e/presubmit_test.go
```diff
@@ -102,6 +102,19 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.PreSubmit(), func(
...
+			BeforeEach(func() {
+				_ = helpers.EnsureControlPlaneMachineSetUpdateStrategy(testFramework, machinev1.RollingUpdate)
+				helpers.MakeControlPlaneMachineSetProviderConfigInvalid(testFramework)
+			})
+
+			AfterEach(func() {
+				helpers.EnsureActiveControlPlaneMachineSet(testFramework)
```
JoelSpeed, line 113: Does this put the provider config back to the original value in the CPMS? I would expect to see the spec reset before the end of the test

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -201,6 +201,10 @@ func (m *openshiftMachineProvider) GetMachineInfos(ctx context.Context, logger l
 	}
 
 	for _, machine := range machineList.Items {
+		if err := m.ensureValidProviderConfig(ctx); err != nil {
```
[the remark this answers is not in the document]
odvarkadaniel: I had it like you are mentioning in the first iteration, but I thought that this might be cleaner to read. I also thought it wouldn't matter, since we would have to do this as well every time we construct the machine provider. If you think it is better the way you say, I will change it.

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -282,6 +286,40 @@ func (m *openshiftMachineProvider) generateMachineInfo(ctx context.Context, logg
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        "tmp-machine-dry-run",
+			Namespace:   m.namespace,
+			Annotations: m.machineTemplate.ObjectMeta.Annotations,
+			Labels:      m.machineTemplate.ObjectMeta.Labels,
+		},
+		Spec: m.machineTemplate.Spec,
+	}
+
+	rawConfig, err := m.providerConfig.RawConfig()
```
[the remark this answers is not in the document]
odvarkadaniel: I investigated this when you mentioned this in Slack, but e.g. on aws, there is the `Placement` correctly in there and makes no issues when dry-run creating nor any webhooks have an issue with that.

test/e2e/presubmit_test.go
odvarkadaniel: Nope. Will add that function.

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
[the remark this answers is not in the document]
JoelSpeed: With this approach, you are constructing the config for each machine, so 3 times every time we construct the machine provider, but, if it's in the main constructor, we only do it once

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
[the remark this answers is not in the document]
JoelSpeed: I'm concerned that this might hold true for one cloud, but not others, also, what if we were to have defaulting vary based on some of the failure domain information, I wonder if we need to handle the defaulting _after_ we inject the failure domain information during the comparison process, WDYT?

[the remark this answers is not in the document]
odvarkadaniel: I get this. It for sure works for azure and aws, but...
If we want to proceed and move this to the creation of the provider config, it gets complicated with indexing the machines to a correct failure domain. I was not sure if I could just tmp assign `failureDomain[0]` since it gets indexed here: https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/main/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go#L250-L263 ?

So I would then opt for dry-run creating the machine while creating the machine provider and the failure domain gets injected as I linked? What do you think about this one?

test/e2e/framework/framework.go
```diff
@@ -224,6 +227,75 @@ func (f *framework) IncreaseProviderSpecInstanceSize(rawProviderSpec *runtime.Ra
...
+}
+
+// updateCredentialsSecretNameAzure updates the credentialSecret.Name field from the ControlPlaneMachineSet.
+func updateCredentialsSecretNameAzure(providerConfig providerconfig.ProviderConfig) (*runtime.RawExtension, error) {
+	cfg := providerConfig.Azure().Config()
+	cfg.CredentialsSecret = nil
+
+	rawBytes, err := json.Marshal(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling aws providerSpec: %w", err)
```
JoelSpeed, line 261:
```suggestion
		return nil, fmt.Errorf("error marshalling azure providerSpec: %w", err)
```

test/e2e/framework/framework.go
```diff
@@ -224,6 +227,75 @@ func (f *framework) IncreaseProviderSpecInstanceSize(rawProviderSpec *runtime.Ra
...
+}
+
+// updateCredentialsSecretNameGCP updates the credentialSecret.Name field from the ControlPlaneMachineSet.
+func updateCredentialsSecretNameGCP(providerConfig providerconfig.ProviderConfig) (*runtime.RawExtension, error) {
+	cfg := providerConfig.GCP().Config()
+	cfg.CredentialsSecret = nil
+
+	rawBytes, err := json.Marshal(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling aws providerSpec: %w", err)
```
JoelSpeed, line 291:
```suggestion
		return nil, fmt.Errorf("error marshalling gcp providerSpec: %w", err)
```

JoelSpeed (commented):
Couple of nits, otherwise LGTM

JoelSpeed:
Known unit flake
/test unit

JoelSpeed:
/approve
/lgtm

Lets see how this goes, I'm still concerned about the potential for different defaults based on the failure domains, which would mean we need to inject the failure domain at a later point, but this is probably ok for now since that would likely mean the data should be within the failure domain itself, so in reality, probably isn't going to be an issue.

The only other issue we may see, is if the defaulting/validation fails because of a missing failure domain value, but we will work that out in short order I'd expect

PR #177, opened 2023-03-13
OCPBUGS-10032: updates: manually requeue when waiting for replicas being ready

damdo:
When a CPMS machine replica is not yet ready
(underlying machine doesn't have .status.phase == Running, or backing node is NotReady)
relying on machine events is not sufficent,
as the machine could already be in Running phase while
the backing node might still be in NotReady condition.
Therefore in order to catch node's NotReady -> Ready changes
(a necessary condition for a CPMS replicas to be considered Ready),
we need to manually requeue.

Surfaced due to the changes introduced in: #171

### DIFF
--- a/pkg/controllers/controlplanemachineset/updates.go
+++ b/pkg/controllers/controlplanemachineset/updates.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"time"
 
 	"github.com/go-logr/logr"
 	machinev1 "github.com/openshift/api/machine/v1"
@@ -151,6 +152,8 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineUpdates(ctx context.C
 //
 // In certain scenarios, there may be indexes with missing Machines. In these circumstances, the update should attempt
 // to create a new Machine to fulfil the requirement of that index.
+//
+//nolint:cyclop
 func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx context.Context, logger logr.Logger, cpms *machinev1.ControlPlaneMachineSet, machineProvider machineproviders.MachineProvider, indexedMachineInfos map[int32][]machineproviders.MachineInfo) (ctrl.Result, error) {
 	logger = logger.WithValues("updateStrategy", cpms.Spec.Strategy.Type)
 
@@ -171,7 +174,7 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
 	// as deletions can continue even if the maxSurge has been already reached.
 	surgeCount := deviseExistingSurge(cpms, sortedIndexedMs)
 
-	var updated bool
+	var updated, shouldRequeue bool
 
 	for _, indexToMachines := range sortedIndexedMs {
 		idx := indexToMachines.index
@@ -183,7 +186,14 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
 			updated = true
 		}
 
-		if done := r.waitForPendingMachines(logger, machines); done {
+		if r.waitForReadyMachine(logger, machines) || r.waitForReplacementMachine(logger, machines) {
+			// Relying on machine events is not sufficient in this case, as the machine could already be in Running phase
+			// while the backing node might still be in NotReady condition. Therefore in order to catch
+			// node's NotReady -> Ready changes (a necessary for CPMS replicas to be Ready), we add a manual requeue.
+			updated, shouldRequeue = true, true
+		}
+
+		if done := r.waitForRemoveMachine(logger, machines); done {
 			updated = true
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -183,7 +186,14 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
 			updated = true
 		}
 
-		if done := r.waitForPendingMachines(logger, machines); done {
+		if r.waitForReadyMachine(logger, machines) || r.waitForReplacementMachine(logger, machines) {
+			// Relying on machine events is not sufficient in this case, as the machine could already be in Running phase
+			// while the backing node might still be in NotReady condition. Therefore in order to catch
+			// node's NotReady -> Ready changes (a necessary for CPMS replicas to be Ready), we add a manual requeue.
+			updated, shouldRequeue = true, true
```
elmiko, line 193: how do we know it needs an update here?

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -233,7 +249,14 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineOnDeleteUpdate(ctx co
 			continue
 		}
 
-		if done := r.waitForPendingMachines(logger, machines); done {
+		if r.waitForReadyMachine(logger, machines) || r.waitForReplacementMachine(logger, machines) {
+			// Relying on machine events is not sufficient in this case, as the machine could already be in Running phase
+			// while the backing node might still be in NotReady condition. Therefore in order to catch
+			// node's NotReady -> Ready changes (a necessary for CPMS replicas to be Ready), we add a manual requeue.
+			updated, shouldRequeue = true, true
```
elmiko, line 256: same question here also

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -248,20 +271,18 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineOnDeleteUpdate(ctx co
...
 }
 
-// check if there are machines in a pending or deleting state and return true or false.
-// this function will take an array of MachineInfos and determine if any of those machines
-// are in a pending or deleting state based on the presence, and state, of other machines
-// in the same array. this function does not block, but gives a signal to the caller about
-// whether there are machines that are still transitioning towards a final state.
-func (r *ControlPlaneMachineSetReconciler) waitForPendingMachines(logger logr.Logger, machines []machineproviders.MachineInfo) bool {
+// waitForReadyMachine checks machines and finds out whether to wait or not for any of them to become ready.
+func (r *ControlPlaneMachineSetReconciler) waitForReadyMachine(logger logr.Logger, machines []machineproviders.MachineInfo) bool {
```
elmiko, line 282: i was under the impression that this function should also be checking the Node's state, given the discussion about Machines being Running while a Node is NotReady. is there a check for node status somewhere else?

elmiko (commented):
code looks mostly good to me, but i have a couple questions

pkg/controllers/controlplanemachineset/updates.go
damdo: `waitForReadyMachine` checks the various `[]machineproviders.MachineInfo` for the `Ready` field.
That field is populated by the machineprovider which [now](https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/171) looks at both Machine phase Running and Node Condition Ready, before setting the field.

pkg/controllers/controlplanemachineset/updates.go
damdo: `update` predates this PR. It is used for deciding whether to log a noop or not:
```
	if !updated {
		logger.V(4).Info(noUpdatesRequired)
	}
```

If we are waiting for something that has its own special log, so we don't want to log the noUpdatesRequired line.

pkg/controllers/controlplanemachineset/updates.go
damdo: See: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/177#discussion_r1134492296

elmiko (commented):
thanks for answering my questions
/lgtm

pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -183,7 +186,14 @@ func (r *ControlPlaneMachineSetReconciler) reconcileMachineRollingUpdate(ctx con
 			updated = true
 		}
 
-		if done := r.waitForPendingMachines(logger, machines); done {
+		if r.waitForReadyMachine(logger, machines) || r.waitForReplacementMachine(logger, machines) {
```
lobziik, line 189: nit: I stumbled on this naming while reading the code... Can we reconsider this naming pattern to something like `needToWaitFor`?

lobziik (commented):
/lgtm

Looks good to me.

I'm not so familiar with this codebase, so I don't feel myself confident to set approval label here.

damdo:
I've spoken with @tjungblu regarding the persistence of the invariant test failure in `e2e-aws-ovn-etcd-scaling`.

Got confirmation that this is not caused by CPMS or this PR but it has been broken for a while[ in other places too](https://github.com/openshift/cluster-etcd-operator/pull/1033).

Proceeding to override this.

PR #179, opened 2023-03-16
E2E pass rate improvement efforts

damdo:
Following some investigations in CPMSO e2e periodics failures I'm proposing the following changes:
- [e2e: periodics: bump wait time for machine replacement creation](https://github.com/openshift/cluster-control-plane-machine-set-operator/commit/c6163f4ad5bb681f2516130aa9a569477cbab48d) 
in my observation it can take a bunch of seconds, 30 seconds, but even more than a
minute to succeed
- [e2e: periodics: RunCheckUnitl don't break e2e if temp errors on get/list](https://github.com/openshift/cluster-control-plane-machine-set-operator/commit/952d284344df163030fad6de07fdc8c89d39d3cb)
if the get()/list() errors it would break the Expect() which would fail
the entire E2Es.
Since we know temporary tcp connection issues with the API server can happen,
we want to make things a bit more robust by ignoring those errors and
checking again later.

TODO:
bug will follow for easy backporting.


test/e2e/helpers/machine.go
```diff
@@ -133,26 +133,36 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
 func EventuallyIndexIsBeingReplaced(ctx context.Context, idx int) bool {
 	controlPlaneMachineSelector := runtimeclient.MatchingLabels(framework.ControlPlaneMachineSetSelectorLabels())
 
-	// Wait up to a minute for the replacement machine to be created.
-	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	// Wait for the replacement machine to be created.
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
lobziik, line 137: IMHO, it's quite significant timeout increasing... Could it mask issues which we would like to know about?

test/e2e/helpers/machine.go
```diff
@@ -133,26 +133,36 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
 		},
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Until function
 			By(fmt.Sprintf("Checking that index %d has 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
-
-			return g.Expect(list()).Should(
+			list, err := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)()
+			if err != nil {
+				// For temporary errors in listing the objects we don't want to break the until condition,
+				// so we return false, which is the standard behaviour for this condition when things haven't settled yet.
```
lobziik, line 161: I suggest to add some logging there. Having error logged will help us to analyse failure if it is occurred.

lobziik (commented):
I'm ok with the relaxing of list results check. But timeout increasing looks quite significant to me, are we sure that this won't mask issues which we would like to know about?

test/e2e/helpers/machine.go
```diff
@@ -133,26 +133,36 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
 	// Use RunCheckUntil to check that the other indexes do not get an additional
 	// replacement while we wait for the desired index to get a replacement.
 	passed := framework.RunCheckUntil(ctx,
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Check function
 			By(fmt.Sprintf("Checking that other indexes (not %d) do not have 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
-
-			return g.Expect(list()).Should(
+			list, err := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)()
+			if err != nil {
```
JoelSpeed, line 146: Is it safe to catch all errors here?

test/e2e/helpers/machine.go
```diff
@@ -133,26 +133,36 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
 				HaveField("Items", WithTransform(extractMachineIndexCounts, Not(HaveKeyWithValue(Not(Equal(idx)), 2)))), fmt.Sprintf("expected not to have 2 replicas for index other than index %d", idx),
 			)
 		},
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Until function
 			By(fmt.Sprintf("Checking that index %d has 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
-
-			return g.Expect(list()).Should(
+			list, err := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)()
+			if err != nil {
```
JoelSpeed, line 159: Likewise, are we sure we want all errors?

test/e2e/helpers/machine.go
```diff
@@ -133,26 +133,36 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
 
 	// Use RunCheckUntil to check that the other indexes do not get an additional
 	// replacement while we wait for the desired index to get a replacement.
 	passed := framework.RunCheckUntil(ctx,
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Check function
 			By(fmt.Sprintf("Checking that other indexes (not %d) do not have 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
-
-			return g.Expect(list()).Should(
+			list, err := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)()
```
JoelSpeed, line 145: Given we are no longer gaining benefit from `komega` here, would it make sense to unwrap this and use the client directly?

test/e2e/helpers/machine.go
```diff
@@ -133,26 +133,36 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
+			return g.Expect(list).Should(
 				HaveField("Items", WithTransform(extractMachineIndexCounts, Not(HaveKeyWithValue(Not(Equal(idx)), 2)))), fmt.Sprintf("expected not to have 2 replicas for index other than index %d", idx),
 			)
 		},
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Until function
 			By(fmt.Sprintf("Checking that index %d has 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
-
-			return g.Expect(list()).Should(
+			list, err := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)()
```
JoelSpeed, line 158: Given we are no longer gaining benefit from `komega` here, would it make sense to unwrap this and use the client directly?

JoelSpeed (commented):
What happens if the check API request fails but the until API request succeeds, is there a way that this could cause an issue where we get a false positive?

test/e2e/helpers/machine.go
damdo: See improvement in: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/179/files#diff-8ed26fa05dca8803c0c37a9d1231d816a93b736abd14cb65905a4856f6af6ceeR510-R525

test/e2e/helpers/machine.go
damdo: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/179#discussion_r1157041968

test/e2e/helpers/machine.go
damdo: Done

test/e2e/helpers/machine.go
damdo: Done

JoelSpeed (commented):
Will the retryable API errors ever be visible to us? It doesn't look like they are being logged at all. Also, if it's not a retryable API error, we break out of the loop, but won't get the error out? I think we somehow need to report back the error.

Perhaps instead we do something like, then if the transform and boolean don't match, it will at least print out the error

```
return g.Expect(err).Should(WithTransform(isRetryableError, BeTrue()))
```

damdo:
Failed on install
/test e2e-aws-operator

test/e2e/helpers/machine.go
```diff
@@ -130,29 +132,42 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Check function
 			By(fmt.Sprintf("Checking that other indexes (not %d) do not have 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
 
-			return g.Expect(list()).Should(
+			list := &machinev1beta1.MachineList{}
+			if err := k8sClient.List(ctx, list, controlPlaneMachineSelector); err != nil {
+				// For temporary errors in listing objects we don't want to break this check,
+				// so we return happy and retry at the next check.
+				return g.Expect(err).Should(WithTransform(isRetryableAPIError(err), BeTrue()))
```
JoelSpeed, line 153: Gomega should apply the transform if you pass a function in
```suggestion
				return g.Expect(err).Should(WithTransform(isRetryableAPIError, BeTrue()))
```

test/e2e/helpers/machine.go
```diff
@@ -130,29 +132,42 @@ func CheckControlPlaneMachineOnDeleteReplacement(testFramework framework.Framewo
...
 		func(_ context.Context, g framework.GomegaAssertions) bool { // Until function
 			By(fmt.Sprintf("Checking that index %d has 2 replicas", idx))
-			list := komega.ObjectList(&machinev1beta1.MachineList{}, controlPlaneMachineSelector)
 
-			return g.Expect(list()).Should(
+			list := &machinev1beta1.MachineList{}
+			if err := k8sClient.List(ctx, list, controlPlaneMachineSelector); err != nil {
+				// For temporary errors in listing the objects we don't want to break the until condition,
+				// so we return false, which is the standard behaviour for this condition when things haven't settled yet.
+				return !g.Expect(err).Should(WithTransform(isRetryableAPIError(err), BeTrue()))
```
JoelSpeed, line 167:
```suggestion
				return g.Expect(err).ShouldNot(WithTransform(isRetryableAPIError, BeTrue()))
```

test/e2e/helpers/machine.go
damdo: 👍

JoelSpeed:
/approve
/lgtm

lets see how the tests run

damdo:
Looked into the failures, nothing related to the improvements we are making:
cc. @JoelSpeed 

e2e-aws-operator : some operators didn't recover in time after the rollout (timed out)
e2e-aws-ovn-etcd-scaling: unrelated. worker node should match or exceed machine count
e2e-aws-ovn-serial: unrelated. annotation not found for node patological event
e2e-aws-ovn-upgrade: unrelated. one second too long oauth disruption

PR #185, opened 2023-04-04
OCPBUGS-11389: Use PlatformStatus instead of PlatformSpec to determine platform

cblecker:
When determining the current platform, we should use PlatformStatus instead of PlatformSpec.


damdo (commented):
Uh! Thanks for catching this. Didn't know that field wasn't there since the beginning of OCP 4.
/approve
/assign cc. @JoelSpeed

JoelSpeed:
/lgtm

Thanks @cblecker, great catch 😅

PR #191, opened 2023-04-12
Inject a failure domain into a dry-run created machine

odvarkadaniel:
We are currently encountering issues with GCP e2e failures introduced by #175. We currently can't dry-run create the machine to ensure the validity of a providerSpec due to failure domain not being in place for this machine.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
@@ -100,9 +100,9 @@ func NewMachineProvider(ctx context.Context, logger logr.Logger, cl client.Clien
 		return nil, errEmptyConfig
 	}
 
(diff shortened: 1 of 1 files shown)


JoelSpeed (commented):
Is there any way we can add some unit test/integration test that would cover this to make sure the failure domain is added before we do the dry run create?

odvarkadaniel:
> Is there any way we can add some unit test/integration test that would cover this to make sure the failure domain is added before we do the dry run create?

I don't think that it is necessary to add new unit tests for this as we don't add any new logic to creating the machine infos.
We know that the logic for injecting FDs already works and there are tests for that. If this logic fails, we will never get to the stuff that is added in this PR.
Other than that, the original PR for this issue has added a new integration test to check the correctness. We can also see that the GCP integration is passing now.

JoelSpeed:
/approve
/lgtm

Ok, I'll accept that for now, but we may want to improve it in the future, I'm skeptical

PR #199, opened 2023-04-19
OCPBUGS-12240: Run `make crds-sync`

pierreprinetti:
I am not sure what I'm doing, but I suspect that either this is the right thing to do, or the tooling needs an update.

I am only running this on the main branch to get a sane diff from another PR where this command was run.

Please advise!


JoelSpeed:
/approve
/lgtm

Will look into why this isn't being caught by CI

PR #201, opened 2023-04-21
docs: Off-by-one error

stephenfin:
Looks like we introduced a new value here in the past (commit 7961d145). Fix numbering, simplifying things somewhat in the process.


damdo:
Only a docs change that doesn't involve units.
Which are flaking.
/override ci/prow/unit

PR #207, opened 2023-05-17
OCPBUGS-7516: fix double machine creation on stale cache

damdo:
Previously, for OCPBUGS-4297 (https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/149), we included a check on whether a replacement machine already existed and was not previously detected due to potential stale cache. The check works, and correctly reports a log when a replacement machine is already present, but we actually forgot to include a return, to stop it from creating the actual machine.


JoelSpeed:
Are there any tests that can be added to prove that this is working?

damdo:
Tests now included.

PR #214, opened 2023-06-02
OCPBUGS-13205: Check for unknown fields in the provider spec

odvarkadaniel:
This PR makes sure that there are no unknown fields in the provider spec. If there are any unknown fields, we log a warning to the user with the unknown field inside the error - as far as I know, there isn't a way to find all the unknown fields at once? We also needed to convert the json bytes to yaml in order for this to work - otherwise the CPMS webhook gets mad.
This change needs to be done to all the providers that are supported by the CPMS.


pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/gcp.go
```diff
@@ -59,8 +61,26 @@ func (g GCPProviderConfig) Config() machinev1beta1.GCPMachineProviderSpec {
 // It should return an error if the provided RawExtension does not represent a GCPProviderConfig.
 func newGCPProviderConfig(raw *runtime.RawExtension) (ProviderConfig, error) {
 	var gcpMachineProviderSpec machinev1beta1.GCPMachineProviderSpec
-	if err := json.Unmarshal(raw.Raw, &gcpMachineProviderSpec); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal GCP provider config: %w", err)
+
+	yamlRaw, err := yaml.JSONToYAML(raw.Raw)
```
JoelSpeed, line 65: What happens if you directly put the `raw.Raw` into `yaml.UnmarshalStrict`? Since json is a subset of YAML I'd have expected it to be ok with that

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/gcp.go
```diff
@@ -59,8 +61,26 @@ func (g GCPProviderConfig) Config() machinev1beta1.GCPMachineProviderSpec {
...
+		if err := json.Unmarshal(raw.Raw, &gcpMachineProviderSpec); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal GCP provider config: %w", err)
+		}
+	}
+
+	// If the gcp provider spec contains an unknown field, we want to log a warning to the user
+	// instead of just omitting the unknown field. This also catches only the first unknown field
+	// in the provider spec. In order to not break any live clusters, we keep the original json
+	// style of unmarshaling if the strict version fails.
+	if err := yaml.UnmarshalStrict(yamlRaw, &gcpMachineProviderSpec); err != nil {
```
JoelSpeed, line 78: Wonder if we can make this unmarshal logic a bit less GCP specific and create a helper function that would easily be used across multiple providers?

odvarkadaniel: Probably easily achievable, will do that.

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/gcp.go
odvarkadaniel: I tried that originally, but I wasn't able to even update the CPMS at all. The strict version of Unmarshal was having issue with multiple fields in the CPMS (mainly gcpMetadata, which weren't present at all in the first place). The way it works is that the rawConfig (originally in json) gets strictly converted to yaml and then back to json in the end. As you are mentioning, it would make sense for the input to be json and the function still working as expected, but not in my case. I will try again just to make sure.

odvarkadaniel:
All the comments are addressed, ready for review.

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/providerconfig.go
```diff
@@ -366,3 +369,22 @@ func ExtractFailureDomainsFromMachineSets(machineSets []machinev1beta1.MachineSe
...
+// checkForUnknownFieldsInProviderSpecAndUnmarshal tries to unmarshal content into a platform specific provider spec
+// and detect invalid fields.
+//
+// If the provider spec contains an unknown field, we want to log a warning to the user
+// instead of just omitting the unknown field. This also catches only the first unknown field
+// in the provider spec. In order to not break any live clusters, we keep the original json
+// style of unmarshaling if the strict version fails.
+func checkForUnknownFieldsInProviderSpecAndUnmarshal(raw *runtime.RawExtension, platformProviderSpec interface{}) error {
+	if err := yaml.UnmarshalStrict(raw.Raw, platformProviderSpec); err != nil {
+		klog.Warning("failed to strictly unmarshal provider config due to unknown field: %w", err)
```
JoelSpeed, line 382: Normally in CPMS we pass a logger into the function, that would probably add a lot of scope creep to this PR, could you investigate what it would mean to pass a `logr.Logger` through to this function?

JoelSpeed:
Changes look good, but will need to rebase and review the linter issues

odvarkadaniel:
> Changes look good, but will need to rebase and review the linter issues

I can either surpress the linter by `nolint:dupl` or refactor the functions for gcp, aws and azure in controlplanemachinesetgenerator:
1. Have a function `buildFailureDomains(...)` instead of platform specific functions, so we will take care of implementing these functions in `utils.go`.
2. As the code below is the only platform specific code, we can have a switch in `buildFailureDomains(...)` and have function that appends the FDs in a platform specific function.
```
gcpFailureDomains := []machinev1.GCPFailureDomain{}
	for _, fd := range failureDomains.List() {
		gcpFailureDomains = append(gcpFailureDomains, fd.GCP())
	}

	cpmsFailureDomain := machinev1.FailureDomains{
		GCP:      &gcpFailureDomains,
		Platform: configv1.GCPPlatformType,
	}
```

Changing it here would mean that the OpenStack PR should also refactor this part to correspond with the other providers.
@JoelSpeed

PR #195, opened 2023-04-14
OSASINFRA-3133 - OpenStack support

EmilienM:
OpenStack support in CPMS.

PR #217, opened 2023-06-15
OSASINFRA-3180: openstack: volume type in failure domain

pierreprinetti:
Depends on #195 
Depends on https://github.com/openshift/api/pull/1496

### DIFF
--- a/docs/user/failure-domains.md
+++ b/docs/user/failure-domains.md
@@ -86,8 +86,9 @@ An Azure failure domain will look something like the example below:
 
 ## OpenStack
 
-On OpenStack, the failure domains represented in the control plane machine set can be considered analogous to the
-OpenStack availability zones (for Nova and Cinder).
+On OpenStack, the failure domains represented in the control plane machine set
+include the OpenStack Nova availability zone for instance placement, as well as
+the Cinder availability zone and the volume type for root volume placement.
 
 > OpenStack Availability Zones are an end-user visible logical abstraction for partitioning an OpenStack cloud without
 > knowing the physical infrastructure. They are used to partition a cloud on arbitrary factors, such as location (country, datacenter, rack),
@@ -103,4 +104,5 @@ An OpenStack failure domain will look something like the example below:
 - availabilityZone: "<nova availability zone>"
   rootVolume:
     availabilityZone: "<cinder availability zone>"
+    volumeType: "<cinder volume type>"
 ```
--- a/docs/user/installation.md
+++ b/docs/user/installation.md
@@ -241,8 +241,10 @@ plane machines and you should populate this on both the Machine and the ControlP
 
 #### Configuring a control plane machine set on OpenStack
 
-Two fields are supported for now: `availabilityZone` (instance AZ) and `rootVolume.availabilityZone` (root volume AZ).
-Gather the existing control plane machines and note the value of the zones of each if they exist.
+The OpenStack failureDomain configuration supports three fields:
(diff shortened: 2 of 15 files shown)


EmilienM:
Nice start @pierreprinetti - I'm thinking, didn't you have to patch the testutils as well for the tests to cover volumeType?

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/openstack.go
```diff
@@ -52,20 +58,22 @@ func (a OpenStackProviderConfig) InjectFailureDomain(fd machinev1.OpenStackFailu
...
+	var failureDomainRootVolume *machinev1.RootVolume
+	if a.providerConfig.RootVolume != nil {
+		failureDomainRootVolume = &machinev1.RootVolume{}
+
+		if az := a.providerConfig.RootVolume.Zone; az != "" {
+			failureDomainRootVolume.AvailabilityZone = az
+		}
+
+		if vt := a.providerConfig.RootVolume.VolumeType; vt != "" {
+			failureDomainRootVolume.VolumeType = vt
```
EmilienM, line 70: here I wonder if we want to add a `else` statement, like:

```golang
} else {
  failureDomainRootVolume = nil
}
```

Because `machinev1.RootVolume{}` is not nil, it has to have at least one property defined:

```golang
// +kubebuilder:validation:MinProperties:=1
type RootVolume struct {
  ...
}
```

I've tried added it, the unit tests still fail, I'm investigating.

EmilienM: I pushed a change but now the unit tests fail differently...

pierreprinetti: The only logical difference your patch is adding, is that it considers the case where `a.providerConfig.RootVolume` is non-nil but empty. Since this causes a change in the tests, it means that in the CPMS generation I have introduced the creation of an empty rootVolume: that's what we should probably remove.

EmilienM:
/test unit
sounds like a flake

docs/user/failure-domains.md
```diff
@@ -86,8 +86,9 @@ An Azure failure domain will look something like the example below:
 
 ## OpenStack
 
-On OpenStack, the failure domains represented in the control plane machine set can be considered analogous to the
-OpenStack availability zones (for Nova and Cinder).
+On OpenStack, the failure domains represented in the control plane machine
```
odvarkadaniel, line 89:
```suggestion
On OpenStack, the failure domains represented in the control plane machine set
```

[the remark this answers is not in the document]
EmilienM: :heavy_check_mark:

docs/user/installation.md
```diff
@@ -253,9 +255,11 @@ For each AZ you have in the cluster, configure a failure domain like below:
 - availabilityZone: "<nova availability zone>"
   rootVolume:
     availabilityZone: "<cinder availability zone>"
+    volumeType: "<cinder volume type>"
 ```
 
-With these zones, the complete `failureDomains` (4 and 5) on the example above should look something like below:
+OpenStack failure domains can't be empty, however each individual property is optional.
```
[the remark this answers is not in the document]
EmilienM: :heavy_check_mark:

docs/user/failure-domains.md
EmilienM: :heavy_check_mark:

PR #220, opened 2023-06-21
OCPBUGS-15338: Fix flaking machine mapping testing

JoelSpeed:
After some debugging on this test, I realised that there was an issue with how we were calculating the mapping and then the machine info.

The mapping is inherently tied to the machine info, since some characteristics of the Machines can influence the mapping. So, logically, the mapping and the machine info should be tied to one another.

What was happening when we were seeing the flakes, is that the controller started a reconcile, calculated a mapping, and _then the machine got marked for deletion_. Because it happened after the mapping, but before `GetMachineInfo` was called, the machine info included the deleting machine, triggering a create on the now stale mapping.

This PR resolves that issue by tying the mapping and machine data together. They should only ever be updated together.
So now, when we create a new provider using `WithClient`, it updates both at the same time. It also means after the initial bootstrap, there are no calls to list machines in `GetMachineInfo`, so the machines it is working with and the mapping should be consistent.

I have also decided I don't really like the name `WithClient` anymore, we should rename it to include some sort of connotation for updating the Machine information that's cached, any suggestions?


pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -51,21 +46,21 @@ var (
...
-	machineMapping, deletingIndexes, err := createMachineMapping(ctx, logger, cl, cpms)
+	machineMapping, deletingIndexes, err := createMachineMapping(logger, machineList)
 	if err != nil {
 		return nil, fmt.Errorf("could not construct machine mapping: %w", err)
 	}
 
 	failureDomainsSet := failuredomain.NewSet(failureDomains...)
 
-	baseMapping, err := createBaseFailureDomainMapping(cpms, failureDomainsSet.List(), machineMapping)
+	baseMapping, err := createBaseFailureDomainMapping(replicas, failureDomainsSet.List(), machineMapping)
```
odvarkadaniel, line 63: Do we really need to pass `replicas`? Wouldn't `len(machineList)` be enough?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -51,21 +46,21 @@ var (
 // to by external code to create new Machines in the same failure domain. It should start with a basic mapping and
 // then use existing Machine information to map failure domains, if possible, so that the Machine names match the
 // index of the failure domain in which they currently reside.
-func mapMachineIndexesToFailureDomains(ctx context.Context, logger logr.Logger, cl client.Client, cpms *machinev1.ControlPlaneMachineSet, failureDomains []failuredomain.FailureDomain) (map[int32]failuredomain.FailureDomain, error) {
+func mapMachineIndexesToFailureDomains(logger logr.Logger, machineList []machinev1beta1.Machine, replicas int32, failureDomains []failuredomain.FailureDomain) (map[int32]failuredomain.FailureDomain, error) {
```
odvarkadaniel, line 49:
```suggestion
func mapMachineIndexesToFailureDomains(logger logr.Logger, machines []machinev1beta1.Machine, replicas int32, failureDomains []failuredomain.FailureDomain) (map[int32]failuredomain.FailureDomain, error) {
```
Confusing with `machineList` type.

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -128,36 +123,26 @@ func createBaseFailureDomainMapping(cpms *machinev1.ControlPlaneMachineSet, fail
...
 
-	deletingIndexes := listDeletingIndexes(machineList.Items)
+	deletingIndexes := listDeletingIndexes(machineList)
 
 	return mapping, deletingIndexes, nil
 }
 
 // mapIndexesToFailureDomainsForMachines creates an index to failure domain mapping for machine in the list.
-func mapIndexesToFailureDomainsForMachines(logger logr.Logger, machineList *machinev1beta1.MachineList) (map[int32]failuredomain.FailureDomain, error) {
+func mapIndexesToFailureDomainsForMachines(logger logr.Logger, machineList []machinev1beta1.Machine) (map[int32]failuredomain.FailureDomain, error) {
```
odvarkadaniel, line 138:
```suggestion
func mapIndexesToFailureDomainsForMachines(logger logr.Logger, machines []machinev1beta1.Machine) (map[int32]failuredomain.FailureDomain, error) {
```
Same as above.

damdo (commented):
Thanks for digging into this, the fix implementation looks reasonable to me.
/approve

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
JoelSpeed: The number of machines doesn't always equal the number of replicas in the cluster, so yes, there are test cases that would fail if you switched this over to len(machineList)

EmilienM:
FYI `e2e-openstack-operator` can safely be ignored. I'll start passing once https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/195 is merged.

damdo:
We would probably want to backport this to 4.13/12

JoelSpeed:
If it's not going to pick cleanly, I might suggest we don't cherry-pick. It's not a major issue and only happens in a small select race between condition. We don't expect the balancing on a cluster to change frequently so I suspect most end users wouldn't actually hit this issue in production

PR #224, opened 2023-07-06
e2e: OpenStack CPMS created automatically

EmilienM:
We manage the OpenStack CPMS via the installer, so we can switch the e2e
in Full, so the framework won't create it for us.

Depends-on: https://github.com/openshift/installer/pull/7280


EmilienM:
/cc pierreprinetti 

since the installer PR is not merged yet, the `e2e-openstack-operator` should fail. It's a good first test to make sure it actually fails, and once we have the installer patch we'll retest and make sure the test passes.

EmilienM:
I don't understand how `e2e-openstack-operator` could pass, I'll investigate a bit.

EmilienM:
@damdo @JoelSpeed for review please

PR #225, opened 2023-07-12
 webhook/cpms: add failure domain validation to OpenStack

EmilienM:
# High level change

* Add a new `checkOpenShiftFailureDomains` function to validate the
  specified failure domains.
* Use this new function on the OpenStack platform to validate two
  things:
  * That rootVolume has an AZ when needed
  * That rootVolume always has a volumeType.
* Add unit test coverage for the OpenStack checks.

# rootVolume `volumeType` validation 

## Context

This PR has an impact for the following environment:

* OpenShift on OpenStack IPI
* The control plane machines are using root volumes
* CPMS is enabled (will be the default in 4.14)

## Facts

Historically, the installer requires a volume type to be specified when deploying a control plane with root volumes:
https://github.com/openshift/installer/blob/ff1f4528081b599790224fce2b0d8c62ebc28014/pkg/asset/installconfig/openstack/validation/machinepool.go#L94-L97

While the `volumeType` is technically [not required](https://github.com/openshift/api/blob/81d582da354baeaa7755e0da51162cf3e8fd5dd5/machine/v1alpha1/types_openstack.go#L356-L358) in `rootVolume` v1alpha1 type, it has [always](https://github.com/openshift/installer/commit/30c95315ac7f01d1cc48a9e33d918b4ee9fca25f#diff-af03f0af30fa6f0ce4070327a45cc60259155197e00fed577d08ca96ccc9a32dR15-R17) been required by the Installer.

So the machines created on day 1 by the installer with a `rootVolume` will **always** have a defined `volumeType`.

## Problem

In the initial implementation of CPMS, we have marked the `volumeType` field as optional, and it's not clear what volume type would it use if a failure domain would exist with a `rootvolume` but no `volumeType`.
Can we safely rely on the default Cinder volume type? Should it be taken from the `providerSpec`? This doesn't align with the rest of CPMS where anything "failure domain related" should live in the Failure Domain API and not in the `providerSpec`.

## Proposed solution

It is proposed to require `volumeType` in a Failure Domain with a `rootVolume`, for multiple reasons:

* To be aligned with what the installer generates on day 1. Remember that machines with a `rootVolume` will **always** have a defined `volumeType`.
* Relying on the default cinder type is risky. For example, the default storage type configured in OpenStack Cinder might not be available or not exist anymore.
  Also, in the context of Availablity zones, it's possible that the default storage type is not available for a given AZ.
  So we think that it's better to be explict than make a decision for the user which might lead to deployment or upgrade errors.
* To not have to deal with a default value in the CPMS `providerSpec` which I don't think is the way CPMS was designed to be working. As far as I understand, the fields related to a failure domain should not go
  in `providerSpec`.

# rootVolume `AvailabilityZone` validation

## Problem

When a machine is created with a compute availability zone (defined via `mpool.zones`) and a storage root volume (defined as `mpool.rootVolume`) and that `rootVolume` has no specified `zones`, CAPO will use the compute AZ for the volume AZ.
This can be problematic if the AZ doesn't exist in Cinder.

Source:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9d183bd479fe9aed4f6e7ac3d5eee46681c518e7/pkg/cloud/services/compute/instance.go#L439-L442

```golang
func (s *Service) getOrCreateRootVolume(eventObject runtime.Object, instanceSpec *InstanceSpec, imageID string) (*volumes.Volume, error) {

(...)

        availabilityZone := instanceSpec.FailureDomain
        if rootVolume.AvailabilityZone != "" {
                availabilityZone = rootVolume.AvailabilityZone
        }

(...)
```

## Proposed solution

If a compute AZ is provided alongside with a root volume, we now require
the root volume to have an AZ, so we force the user to make a choice on
which AZ the root volume is deployed on.

We are also enforcing it via CEL validation in https://github.com/openshift/api/pull/1521 and in the installer via https://github.com/openshift/installer/pull/7309

### DIFF
--- a/pkg/webhooks/controlplanemachineset/webhooks.go
+++ b/pkg/webhooks/controlplanemachineset/webhooks.go
@@ -373,6 +373,8 @@ func checkOpenShiftFailureDomainsMatchMachines(logger logr.Logger, parentPath *f
 		return append(errs, field.Invalid(parentPath, failureDomains, fmt.Sprintf("error getting failure domains from control plane machine set machine template: %v", err)))
 	}
 
+	errs = append(errs, checkOpenShiftFailureDomains(specifiedFailureDomains)...)
+
 	// Failure domains used by control plane machines but not specified in the control plane machine set
 	if missingFailureDomains := missingFailureDomains(machineFailureDomains, specifiedFailureDomains); len(missingFailureDomains) > 0 {
 		errs = append(errs, field.Forbidden(parentPath, fmt.Sprintf("control plane machines are using unspecified failure domain(s) %s", missingFailureDomains)))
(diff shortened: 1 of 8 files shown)


EmilienM:
The unit tests don't work yet, I'm trying to figure out why.

EmilienM:
In the context of https://github.com/openshift/api/pull/1520

I've been trying to add some validations here (outside of the webhook). The tests aren't passing yet and I don't know why. Also I'm not sure if I'm on the right path. Joel asked to look at the webhook but I'm interested to validate a FailureDomain and not the providerSpec value from the CPMS.
Feedback highly welcome.

cc JoelSpeed pierreprinetti

EmilienM:
@damdo @JoelSpeed I'll like early feedback on this one before I continue down the hole to update the rest of the unit tests for OpenStack, with proper rootVolume properties.
Does it sound reasonable and the right place to do these things?

@pierreprinetti when you return from holidays, maybe you can provide some feedback on this proposal (replacing the CEL validation in API initially proposed).

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -388,6 +390,39 @@ func checkOpenShiftFailureDomainsMatchMachines(logger logr.Logger, parentPath *f
...
+			errs = append(errs, fmt.Errorf("rootVolume availabilityZone must be specified when availabilityZone is specified")) //nolint:goerr113
+		}
+
+		if failureDomain.RootVolume.VolumeType == "" {
+			errs = append(errs, fmt.Errorf("rootVolume volumeType must be specified")) //nolint:goerr113
+		}
+	}
+
+	return errs
+}
```
JoelSpeed, line 424: I'm surprised we can't achieve this with CEL validation?

EmilienM:
Maybe I'm over-thinking, but for now I'll take a break from this PR and wait for Pierre to come back from vacation. We'll discuss whether we really need to validate the `rootVolume.volumeType`, while the field should be set when the machines are created by the installer.
I really wanted to add safeguards so we don't let customers do exotic things; but it seems like we need further discussion on the how.

I'm personally convinced that we should solve this one via the CEL validations (and not here) for 2 reasons:
* We haven't reached GA yet, so we can make any change we want in the API with no impact
* If a failure domain with a rootVolume is being updated to have a volumeType, there is very little (none) chance that the control plane would be rolled out because of a new providerSpec outcome, since the volumeType should already be present since we control that in the installer. The only chance for that to happen is if a customer would manually edit a control plane Machine to remove the volumeType (why?) and then indeed the machine would be rolled out on the new failure domain.

EmilienM:
the volumeType validation has moved to CEL test in API, and I'm updating CCPMSO here: https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/231

PR #226, opened 2023-07-14
vendoring: update openshift/api

EmilienM:
Update CPMS with a new openshift/api, which includes changes in the FailureDomains.


damdo (commented):
/approve
/lgtm

PR #229, opened 2023-07-27
OCPCLOUD-2130: Add subnet to Azure FD, fix for optional fields in FD

RadekManak:
This PR 

1. Adds a subnet field to azure failure domain. 
2. Updates the injectFailureDomain functions to keep values from template when failure domain has zero value field. (Like openstack does already). 
3. Updates the failuredomain to machine mapping function to handle cases where part of failure domain is provided by template and other is different across failure domains.


pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -65,6 +65,9 @@ type FailureDomain interface {
 
 	// Equal compares the underlying failure domain.
 	Equal(other FailureDomain) bool
+
+	// RemoveShared Sets field to zero value if they have same value as other FailureDomain field.
+	RemoveTemplateFailureDomainFields(templateFailureDomain FailureDomain) (FailureDomain, error)
```
JoelSpeed, line 70: Comment doesn't match field name

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -142,6 +145,91 @@ func (f failureDomain) Equal(other FailureDomain) bool {
 	return true
 }
 
+// TODO ERROR handling
```
JoelSpeed, line 148: Godoc update needed

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -142,6 +145,91 @@ func (f failureDomain) Equal(other FailureDomain) bool {
...
+func (f failureDomain) removeSharedOpenStack(template machinev1.OpenStackFailureDomain) (FailureDomain, error) {
+	cleanedFD, err := zeroMatchingFields(f.aws, template)
+	if err != nil {
+		return nil, fmt.Errorf("zeroMatchingFields failure: %w", err)
+	}
+	return NewOpenStackFailureDomain(cleanedFD.(machinev1.OpenStackFailureDomain)), nil
+}
+
+// Generic function to set fields of struct a to zero value if they equal b struct field value.
+func zeroMatchingFields(a, b interface{}) (interface{}, error) {
```
JoelSpeed, line 205: I wonder if this needs to be able to recurse? Thinking about AWS, the subnet you would want to replace the entire thing, it is atomic, but maybe on OpenStack the rootVolume there is not atomic? WDYT? Perhaps we need a way to identify if they should be atomic or not.

I think adding some unit tests for this will probably answer this question

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping.go
```diff
@@ -425,10 +441,13 @@ func swapIndexes(mapping map[int32]failuredomain.FailureDomain, x, y int32) {
...
 
 // copyMapping creates a new map with a copy of the keys and values from the source mapping.
-func copyMapping(mapping map[int32]failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
+func copyMapping(mapping map[int32]failuredomain.FailureDomain, templateFailureDomain failuredomain.FailureDomain) map[int32]failuredomain.FailureDomain {
 	out := make(map[int32]failuredomain.FailureDomain)
 
 	for idx, val := range mapping {
+		if templateFailureDomain != nil {
+			val, _ = val.RemoveTemplateFailureDomainFields(templateFailureDomain)
+		}
```
JoelSpeed, line 450: Do you need to do it here as well as in the `createBaseFailureDomainMapping` and `mapMachineIndexesToFailureDomains` functions? Seems like you're maybe running through it twice?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -1548,7 +1566,29 @@ var _ = Describe("Failure Domain Mapping", func() {
 					0: failuredomain.NewAWSFailureDomain(usEast1bFailureDomainBuilder.Build()),
 					1: failuredomain.NewAWSFailureDomain(usEast1cFailureDomainBuilder.Build()),
 					2: failuredomain.NewAWSFailureDomain(usEast1dFailureDomainBuilder.Build()),
-					3: failuredomain.NewAWSFailureDomain(usEast1aFailureDomainBuilder.Build()),
+					3: failuredomain.NewAWSFailureDomain(usEast1eFailureDomainBuilder.Build()),
+					4: failuredomain.NewAWSFailureDomain(usEast1aFailureDomainBuilder.Build()),
+				},
+			}),
+			Entry("with five failure domains and three machines, should prioritise machine domains; a is from template", createBaseMappingTableInput{
```
JoelSpeed, line 1573: Template in these is normally empty right?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -1233,14 +1246,19 @@ var _ = Describe("Failure Domain Mapping", func() {
...
 			if in.expectedError != nil {
 				Expect(err).To(MatchError(in.expectedError))
 			} else {
 				Expect(err).ToNot(HaveOccurred())
 			}
 
+			for i, fd := range in.expectedMapping {
+				fd, _ := fd.RemoveTemplateFailureDomainFields(templateFailureDomain)
+				in.expectedMapping[i] = fd
+			}
```
JoelSpeed, line 1261: Does the expected mapping need pruning? Perhaps we should not do this and make sure the tests expected mapping provides the actual output?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -234,20 +242,25 @@ var _ = Describe("Failure Domain Mapping", func() {
 				machines = append(machines, *machine)
 			}
 
-			mapping, err := mapMachineIndexesToFailureDomains(logger.Logger(), machines, in.replicas, failureDomains)
+			mapping, err := mapMachineIndexesToFailureDomains(logger.Logger(), machines, in.replicas, failureDomains, templateFailureDomain)
```
JoelSpeed, line 245: Template failure domain is currently from the global right? Might be good to have the global one as empty  and then be able to override individually in a test? WDYT?

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -234,20 +242,25 @@ var _ = Describe("Failure Domain Mapping", func() {
...
 			if in.expectedError != nil {
 				Expect(err).To(MatchError(in.expectedError))
 			} else {
 				Expect(err).ToNot(HaveOccurred())
 			}
 
+			for i, fd := range in.expectedMapping {
+				fd, _ := fd.RemoveTemplateFailureDomainFields(templateFailureDomain)
+				in.expectedMapping[i] = fd
+			}
```
JoelSpeed, line 255: Suspect this isn't what we want to do and would be better to have the expected as the actual output

RadekManak:
Still WIP, it compiles now, a looks like it works. There are still some issues in tests. I'd like to avoid changing the AWS resourcebuilders. We are probably the only consumers right now. But it is technically an API now. Also, I may be hiding an actual issue by changing the constructor.

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain_test.go
```diff
@@ -576,4 +577,245 @@ var _ = Describe("FailureDomains", func() {
 
 	})
 
+	Context("RemoveTemplateFailureDomainFields", func() {
+		type testCase struct {
+			FailureDomain         FailureDomain
+			TemplateFailureDomain FailureDomain
+			ExpectedResult        FailureDomain
+			ExpectedError         error
```
JoelSpeed, line 585: These should all be private fields

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -31,15 +33,20 @@ import (
 	corev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/core/v1"
 	machinev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1"
 	machinev1beta1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1beta1"
+	providerConfig "github.com/openshift/cluster-control-plane-machine-set-operator/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-var _ = Describe("Failure Domain Mapping", func() {
+var _ = FDescribe("Failure Domain Mapping", func() {
```
JoelSpeed, line 41: To remove before we merge

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
JoelSpeed, line 245: Should this be coming from the test case?
```suggestion
			mapping, err := mapMachineIndexesToFailureDomains(logger.Logger(), machines, in.replicas, failureDomains, tc.templateFailureDomain)
```

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -234,20 +242,25 @@ var _ = Describe("Failure Domain Mapping", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 			}
 
+			for i, fd := range in.expectedMapping {
+				fd, _ := fd.RemoveTemplateFailureDomainFields(templateFailureDomain)
+				in.expectedMapping[i] = fd
+			}
 			Expect(mapping).To(Equal(in.expectedMapping))
-			Expect(logger.Entries()).To(ConsistOf(in.expectedLogs))
+			//Expect(logger.Entries()).To(ConsistOf(in.expectedLogs))
```
JoelSpeed, line 257: Remove before merge

pkg/machineproviders/providers/openshift/machine/v1beta1/mapping_test.go
```diff
@@ -1790,10 +1808,14 @@ var _ = Describe("Failure Domain Mapping", func() {
...
-				mapping := reconcileMappings(logger.Logger(), in.baseMapping, in.machineMapping, in.deletingIndexes)
+				mapping := reconcileMappings(logger.Logger(), in.baseMapping, in.machineMapping, templateFailureDomain, in.deletingIndexes)
 
+				for i, fd := range in.expectedMapping {
+					fd, _ := fd.RemoveTemplateFailureDomainFields(templateFailureDomain)
+					in.expectedMapping[i] = fd
+				}
 				Expect(mapping).To(Equal(in.expectedMapping))
-				Expect(logger.Entries()).To(Equal(in.expectedLogs))
+				//Expect(logger.Entries()).To(Equal(in.expectedLogs))
```
JoelSpeed, line 1818: Remove before merge

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -201,7 +201,9 @@ func (m *openshiftMachineProvider) updateMachineCache(ctx context.Context, logge
 	}
 
 	// Since the mapping depends on the state of the machines, we must re-map the failure domains if the machines have changed.
-	indexToFailureDomain, err := mapMachineIndexesToFailureDomains(logger, machineList.Items, m.replicas, m.failureDomains)
+	config, err := providerconfig.NewProviderConfigFromMachineSpec(logger, m.machineTemplate.Spec)
```
JoelSpeed, line 204: Err check missing

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -389,7 +404,7 @@ func (m *openshiftMachineProvider) getMachineIndex(logger logr.Logger, machine m
 			"Could not gather Machine Info",
 		)
 
-		return 0, errCouldNotDetermineMachineIndex
+		return 0, fmt.Errorf("%w: %s", errCouldNotDetermineMachineIndex, failureDomain.String())
```
JoelSpeed, line 407: Is this for debugging?

pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/azure_test.go
```diff
@@ -35,9 +35,13 @@ var _ = Describe("Azure Provider Config", func() {
 	zone1 := "1"
 	zone2 := "2"
 
+	subnet_zone1 := "subnet-zone-1"
+	subnet_zone2 := "subnet-zone-2"
```
JoelSpeed, line 39: We use pascalCase, not snake_case

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -240,7 +240,9 @@ func validateOpenShiftMachineV1BetaTemplateOnCreate(logger logr.Logger, parentPa
 	if template.FailureDomains.Platform == "" {
 		errs = append(errs, checkOpenShiftProviderSpecFailureDomainMatchesMachines(logger, parentPath.Child("spec", "providerSpec"), template, machines)...)
 	} else {
-		errs = append(errs, checkOpenShiftFailureDomainsMatchMachines(logger, parentPath.Child("failureDomains"), template.FailureDomains, machines)...)
+		providerConfig, _ := providerconfig.NewProviderConfigFromMachineSpec(logger, template.Spec)
```
JoelSpeed, line 243: Error check?

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -358,17 +360,32 @@ func checkOpenShiftProviderSpecFailureDomainMatchesMachines(logger logr.Logger,
 	return errs
 }
 
+func cleanFailureDomaniArray(failureDomains []failuredomain.FailureDomain, templateFailureDomain failuredomain.FailureDomain) ([]failuredomain.FailureDomain, error) {
+	cleanedFailureDomains := []failuredomain.FailureDomain{}
+	for _, failureDomain := range failureDomains {
+		failureDomain, err := failureDomain.RemoveTemplateFailureDomainFields(templateFailureDomain)
+		if err != nil {
+			return nil, fmt.Errorf("RemoveShared failure: %w", err)
```
JoelSpeed, line 368: Errors cannot start with capitals, and should be in prose, function names like this are not expected

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -65,6 +71,9 @@ type FailureDomain interface {
 
 	// Equal compares the underlying failure domain.
 	Equal(other FailureDomain) bool
+
+	// CombineWithTemplateFailureDomain creates a copy of templateFailureDomain and overrides any set values with the values from the current failure domain.
```
JoelSpeed, line 75: Bit of a nit, but, the behaviour we care about here is ensuring that we have a fully populated failure domain? So perhaps the wording of this should be. Same would go for other functions near here. Could even maybe rename to `EnsureCompleteFailureDomain` or just `Complete`?
```suggestion
	// CombineWithTemplateFailureDomain ensures that any empty value in the failure domain is populated based on the template failure domain, to create a fully specified failure domain.
```

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -142,6 +151,109 @@ func (f failureDomain) Equal(other FailureDomain) bool {
...
+}
+
+func (f failureDomain) combineWithTemplateFailureDomainOpenStack(templateFailureDomain machinev1.OpenStackFailureDomain) FailureDomain {
+	fd := templateFailureDomain.DeepCopy()
+
+	if f.openstack.AvailabilityZone != "" {
+		fd.AvailabilityZone = f.openstack.AvailabilityZone
+	}
+
+	//nolint: nestif
```
JoelSpeed, line 239: What did the linter suggest here?

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -376,7 +383,20 @@ func (m *openshiftMachineProvider) getMachineIndex(logger logr.Logger, machine m
 		return 0, fmt.Errorf("cannot extract failure domain from machine: %w", err)
 	}
 
-	index, indexFound := m.failureDomainToIndex(failureDomain)
+	providerConfig, err := providerconfig.NewProviderConfigFromMachineSpec(logger, m.machineTemplate.Spec)
+	if err != nil {
+		return 0, fmt.Errorf("cannot get provider config for machine: %w", err)
+	}
+	// tamplateFailureDomain is a failure domain from the machine template provider spec.
```
JoelSpeed, line 390: Typo here

pkg/machineproviders/providers/openshift/machine/v1beta1/failuredomain/failuredomain.go
```diff
@@ -142,6 +151,109 @@ func (f failureDomain) Equal(other FailureDomain) bool {
...
+		} else {
+			if f.openstack.RootVolume.AvailabilityZone != "" {
+				fd.RootVolume.AvailabilityZone = f.openstack.RootVolume.AvailabilityZone
+			}
+
+			if f.openstack.RootVolume.VolumeType != "" {
+				fd.RootVolume.VolumeType = f.openstack.RootVolume.VolumeType
+			}
+		}
+	}
```
JoelSpeed, line 252: I think this is equivalent and reduces the nesting a bit, might be able to get rid of that linter marker
```suggestion
	if fd.RootVolume == nil {
		fd.RootVolume = f.openstack.RootVolume
	} else if f.openstack.RootVolume != ni {
		if f.openstack.RootVolume.AvailabilityZone != "" {
			fd.RootVolume.AvailabilityZone = f.openstack.RootVolume.AvailabilityZone
		}
	
		if f.openstack.RootVolume.VolumeType != "" {
			fd.RootVolume.VolumeType = f.openstack.RootVolume.VolumeType
		}
	}
```

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
JoelSpeed: Typo still here

PR #230, opened 2023-08-01
OCPBUGS-18013: Enhance wehbooks to dry run machine creation to validate provider spec

Karthik-K-N:
In machine-api-operators webhook we have various validations defined for machines, Here we are trying to make use of those validation by running a dry-run to create machine with cpms machine spec.

Sample error when try to edit exiting cpms to remove ami id

```
error: controlplanemachinesets.machine.openshift.io "cluster" could not be patched: admission webhook "controlplanemachineset.machine.openshift.io" denied the request: invalid machine spec: admission webhook "validation.machine.machine.openshift.io" denied the request: providerSpec.ami: Required value: expected providerSpec.ami.id to be populated
```

### DIFF
--- a/pkg/webhooks/controlplanemachineset/webhooks.go
+++ b/pkg/webhooks/controlplanemachineset/webhooks.go
@@ -100,7 +100,7 @@ func (r *ControlPlaneMachineSetWebhook) ValidateCreate(ctx context.Context, obj 
 	}
 
 	errs = append(errs, validateMetadata(field.NewPath("metadata"), cpms.ObjectMeta)...)
-	errs = append(errs, validateSpec(r.logger, field.NewPath("spec"), cpms)...)
+	errs = append(errs, r.validateSpec(ctx, field.NewPath("spec"), cpms)...)
 	errs = append(errs, r.validateSpecOnCreate(ctx, field.NewPath("spec"), cpms)...)
 
 	if len(errs) > 0 {
@@ -126,7 +126,7 @@ func (r *ControlPlaneMachineSetWebhook) ValidateUpdate(ctx context.Context, oldO
 	}
 
 	errs = append(errs, validateMetadata(field.NewPath("metadata"), cpms.ObjectMeta)...)
-	errs = append(errs, validateSpec(r.logger, field.NewPath("spec"), cpms)...)
+	errs = append(errs, r.validateSpec(ctx, field.NewPath("spec"), cpms)...)
 
 	if len(errs) > 0 {
 		return warnings, utilerrors.NewAggregate(errs)
@@ -177,16 +177,16 @@ func validateMetadata(parentPath *field.Path, metadata metav1.ObjectMeta) []erro
 }
 
 // validateSpec validates that the spec of the ControlPlaneMachineSet resource is valid.
-func validateSpec(logger logr.Logger, parentPath *field.Path, cpms *machinev1.ControlPlaneMachineSet) []error {
+func (r *ControlPlaneMachineSetWebhook) validateSpec(ctx context.Context, parentPath *field.Path, cpms *machinev1.ControlPlaneMachineSet) []error {
 	errs := []error{}
 
-	errs = append(errs, validateTemplate(logger, parentPath.Child("template"), cpms.Spec.Template, cpms.Spec.Selector)...)
+	errs = append(errs, r.validateTemplate(ctx, cpms.Namespace, parentPath.Child("template"), cpms.Spec.Template, cpms.Spec.Selector)...)
 
 	return errs
 }
 
 // validateTemplate validates the common (on create and update) checks for the ControlPlaneMachineSet template.
-func validateTemplate(logger logr.Logger, parentPath *field.Path, template machinev1.ControlPlaneMachineSetTemplate, selector metav1.LabelSelector) []error {
+func (r *ControlPlaneMachineSetWebhook) validateTemplate(ctx context.Context, namespaceName string, parentPath *field.Path, template machinev1.ControlPlaneMachineSetTemplate, selector metav1.LabelSelector) []error {
 	switch template.MachineType {
 	case machinev1.OpenShiftMachineV1Beta1MachineType:
 		openshiftMachineTemplatePath := parentPath.Child(string(machinev1.OpenShiftMachineV1Beta1MachineType))
@@ -197,7 +197,7 @@ func validateTemplate(logger logr.Logger, parentPath *field.Path, template machi
 			return []error{field.Required(openshiftMachineTemplatePath, fmt.Sprintf("%s is required when machine type is %s", machinev1.OpenShiftMachineV1Beta1MachineType, machinev1.OpenShiftMachineV1Beta1MachineType))}
 		}
 
-		return validateOpenShiftMachineV1BetaTemplate(logger, openshiftMachineTemplatePath, *template.OpenShiftMachineV1Beta1Machine, selector)
+		return r.validateOpenShiftMachineV1BetaTemplate(ctx, namespaceName, openshiftMachineTemplatePath, *template.OpenShiftMachineV1Beta1Machine, selector)
 	default:
 		return []error{field.NotSupported(parentPath.Child("machineType"), template.MachineType, []string{string(machinev1.OpenShiftMachineV1Beta1MachineType)})}
 	}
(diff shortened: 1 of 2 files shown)


pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -223,12 +223,51 @@ func validateTemplateOnCreate(logger logr.Logger, parentPath *field.Path, templa
...
+	errs := []error{}
+	providerConfig, err := providerconfig.NewProviderConfigFromMachineTemplate(r.logger, template)
+	if err != nil {
+		return []error{field.Invalid(parentPath, template.Spec.ProviderSpec.Value, fmt.Sprintf("error determining provider configuration: %s", err))}
+	}
+	templateProviderConfig := providerConfig
 
+	if template.FailureDomains.Platform != "" {
+		templateProviderSpecFailureDomain := templateProviderConfig.ExtractFailureDomain()
+		injectedProviderConfig, err := templateProviderConfig.InjectFailureDomain(templateProviderSpecFailureDomain)
```
JoelSpeed, line 245: You extract the failure domain to inject it back in again? This is going to be a no-op. You need to pull in a failure domain from the `FailureDomains` struct. I think we have examples of doing this generically somewhere in the MachineProvider code already

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -223,12 +223,51 @@ func validateTemplateOnCreate(logger logr.Logger, parentPath *field.Path, templa
...
+		if err != nil {
+			return []error{field.Invalid(parentPath, template.FailureDomains, fmt.Sprintf("error injecting failure domain into provider config: %s", err))}
+		}
+		templateProviderConfig = injectedProviderConfig
+	}
+	dryRunMachine := &machinev1beta1.Machine{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "tmp-machine-dry-run",
+			Namespace:    "openshift-machine-api",
```
JoelSpeed, line 254: You should pull the namespace from the CPMS rather than using a constant `openshift-machine-api` here, was that tried?

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -223,12 +223,51 @@ func validateTemplateOnCreate(logger logr.Logger, parentPath *field.Path, templa
...
+
+	rawConfig, err := templateProviderConfig.RawConfig()
+	if err != nil {
+		return []error{field.Invalid(parentPath, template.Spec.ProviderSpec, fmt.Sprintf("could not fetch raw config from provider config: %s", err))}
+	}
+	dryRunMachine.Spec.ProviderSpec.Value.Raw = rawConfig
+
+	dryRunClient := client.NewDryRunClient(r.client)
+	if err := dryRunClient.Create(ctx, dryRunMachine); err != nil {
+		errs = append(errs, fmt.Errorf("invalid machine spec, could not create the machine: %w", err))
```
JoelSpeed, line 269: I would probably drop the `could not create the machine` part of this, you're revealing inner details that aren't important IMO

pkg/webhooks/controlplanemachineset/webhooks_test.go
```diff
@@ -57,7 +57,11 @@ var _ = Describe("Webhooks", func() {
 		By("Setting up a namespace for the test")
 		ns := corev1resourcebuilder.Namespace().WithGenerateName("control-plane-machine-set-webhook-").Build()
 		Expect(k8sClient.Create(ctx, ns)).To(Succeed())
+		machineNs := corev1resourcebuilder.Namespace().WithName("openshift-machine-api").Build()
+		Expect(k8sClient.Create(ctx, machineNs)).To(Succeed())
+
 		namespaceName = ns.GetName()
+		machineNamespace = machineNs.GetName()
```
JoelSpeed, line 64: There should be no resources in this namespace, all resources are created in the test namespace so I would expect these changes to not be needed

JoelSpeed (commented):
CC @damdo I'd like you to review this as well please

pkg/webhooks/controlplanemachineset/webhooks.go
Karthik-K-N: Intention was to get it from CPMS, did not find a way to fetch it.

Karthik-K-N: Updated accordingly

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -223,11 +223,61 @@ func validateTemplateOnCreate(logger logr.Logger, parentPath *field.Path, templa
...
+
+	templateProviderConfig := providerConfig
+
+	if template.FailureDomains.Platform != "" {
+		failureDomains, err := failuredomain.NewFailureDomains(template.FailureDomains)
+		if err != nil {
+			return []error{field.Invalid(parentPath, template.FailureDomains, fmt.Sprintf("error constructing failure domain config:: %s", err))}
+		}
+
+		injectedProviderConfig, err := templateProviderConfig.InjectFailureDomain(failureDomains[0])
```
Karthik-K-N, line 252: is it safe index the list directly or should we iterate over list to inject the first failure domain?

JoelSpeed: I think lets check at least one failure domain as a separate if clause above, then we can index like this

Karthik-K-N: sure, will handle accordingly

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -223,11 +223,63 @@ func validateTemplateOnCreate(logger logr.Logger, parentPath *field.Path, templa
...
+	if err != nil {
+		return []error{field.Invalid(parentPath, template.Spec.ProviderSpec.Value, fmt.Sprintf("error determining provider configuration: %s", err))}
+	}
+
+	templateProviderConfig := providerConfig
+
+	if template.FailureDomains.Platform != "" {
+		failureDomains, err := failuredomain.NewFailureDomains(template.FailureDomains)
+		if err != nil {
+			return []error{field.Invalid(parentPath, template.FailureDomains, fmt.Sprintf("error constructing failure domain config:: %s", err))}
```
JoelSpeed, line 249: nit shouldn't have double colon
```suggestion
			return []error{field.Invalid(parentPath, template.FailureDomains, fmt.Sprintf("error constructing failure domain config: %s", err))}
```

pkg/webhooks/controlplanemachineset/webhooks.go
Karthik-K-N: Changed the logic to fetch the failure domain.

pkg/webhooks/controlplanemachineset/webhooks.go
Karthik-K-N: Fixed.

damdo (commented):
/lgtm

JoelSpeed:
/approve

We should probably create a bug for this and try to get it merged into 4.14, since the validation may cause issues for new IBM installs with CPMS if we don't right? Or are you happy for this to start in 4.15?

Karthik-K-N:
> /approve
> 
> We should probably create a bug for this and try to get it merged into 4.14, since the validation may cause issues for new IBM installs with CPMS if we don't right? Or are you happy for this to start in 4.15?

I think we should be good to get started with 4.15, Currently only Power VS supports CPMS  in Inactive state.

JoelSpeed:
> I think we should be good to get started with 4.15, Currently only Power VS supports CPMS in Inactive state.

I know that PowerVS only creates the CPMS in inactive state for now, but that doesn't stop a user activating it. You need to think of CPMS as supported in 4.14.

Since this isn't strictly related to PowerVS and affects all platforms, I think it's a separate point though. The sooner that we merge, the fewer people have a chance to break things before we introduce this

Karthik-K-N:
> > I think we should be good to get started with 4.15, Currently only Power VS supports CPMS in Inactive state.
> 
> I know that PowerVS only creates the CPMS in inactive state for now, but that doesn't stop a user activating it. You need to think of CPMS as supported in 4.14.
> 
> Since this isn't strictly related to PowerVS and affects all platforms, I think it's a separate point though. The sooner that we merge, the fewer people have a chance to break things before we introduce this

got the point, Will create a bug to merge it in 4.14. thanks

JoelSpeed:
No, this will merge directly into 4.14 as we haven't branched yet

PR #232, opened 2023-08-09
test/e2e: override `RootVolume.VolumeType` to ""

EmilienM:
`VolumeType` is now part of the Failure Domain so we don't want it in
the CPMS Provider Spec.

This was missed in a previous PR.


EmilienM:
@JoelSpeed any chance we /override here, so we don't need to waste CI

PR #234, opened 2023-08-22
OCPBUGS-17851: Reimplement "OCPCLOUD-2130: Add subnet to Azure FD, fix for optional fields in FD"

RadekManak:
This PR reintroduces #229. The revert of revert cannot apply cleanly, so I've squashed the commit together.

- This PR introduces a 'subnet' field to Azure failure domains.
- It implements functionality where the failure domains field only overrides fields from the template when the fields in the failure domain are not zero.
- Additionally, it implements additional logic to be able to compare failure domains that have some fields inherited from the machine template.
- The 'ControlPlaneMachineSetGenerator' is updated to omit fields that are shared between all machines from the failure domain and generate them in the template instead.


RadekManak:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-azure-sdn-upgrade 10

JoelSpeed:
/hold 

Milind found an issue in the CPMS generation now, @RadekManak PTAL

pkg/controllers/controlplanemachinesetgenerator/azure.go
```diff
@@ -55,14 +58,26 @@ func generateControlPlaneMachineSetAzureSpec(logger logr.Logger, machines []mach
...
+			return nil, fmt.Errorf("failed to extract machine's azure providerSpec: %w", err)
+		}
+
+		providerConfigs = append(providerConfigs, providerconfig)
 	}
 
-	azureProviderSpec := providerConfig.Azure().Config()
-	// Remove field related to the faliure domain.
-	azureProviderSpec.Zone = nil
+	azureProviderSpec := providerConfigs[0].Azure().Config()
```
JoelSpeed, line 72: What if there are no machines? Then this will cause an index out of bounds, consider the UPI cluster case, I think you need a guard to return early without creating a CPMS if there are no machines

RadekManak: The first thing that the ControlPlaneMachineSetGeneratorReconciler does is it gets control plane Machines and exits if we have less than 2. There are more places that assume we have machines. I think it is fine as the function are all private.

RadekManak:
/payload-aggregate periodic-ci-openshift-release-master-ci-4.15-upgrade-from-stable-4.14-e2e-azure-ovn-upgrade 10

RadekManak:
I've updated the description to match current contents of the PR.

/payload-aggregate periodic-ci-openshift-release-master-ci-4.15-upgrade-from-stable-4.14-e2e-azure-ovn-upgrade 10
/hold cancel

PR #235, opened 2023-08-24
OCPBUGS-17055: Address long acquire times during upgrade

odvarkadaniel:


odvarkadaniel:
/hold

Need to test this.

PR #237, opened 2023-08-24
OCPBUGS-18056: Address long acquire times during upgrade

odvarkadaniel:


JoelSpeed:
Please update the PR description with details of the change made and why this is safe to do, what checks were done to make sure that the controller here shuts down gracefully and observes contexts correctly

odvarkadaniel:
> Please update the PR description with details of the change made and why this is safe to do, what checks were done to make sure that the controller here shuts down gracefully and observes contexts correctly

Working on that now.

odvarkadaniel:
This change should lead to shorter times when upgrading OpenShift versions. The leader, with this change, can step down voluntarily when the manager gets shut down.

I think that this change is safe as the binary ends as soon as the manager ends. With this comes, that we need to pass the context correctly throughout the operator, which in this case is also correct.

I have checked that the binary shuts down with a manager end and also checked that context should get passed correctly. With a graceful shutdown of the manager, we should see a significant speed up in release of leader as the new leader doesn't have to wait for `LeaseDuration` time first.

I would be more confident if @sunzhaohua2 could pre-merge test this and verify my thoughts up. Thank you!

sunzhaohua2:
> I would be more confident if @sunzhaohua2 could pre-merge test this and verify my thoughts up.

@odvarkadaniel I added a comment in bug, the time doesn't seem to shorten

damdo (commented):
/lgtm

PR #239, opened 2023-09-05
OCPCLOUD-2167: Test unhealthy node cases

odvarkadaniel:
At the moment, we are missing any cases that would cover unhealthy node cases, such as deleting an instance in the cloud or terminating a kubelet on a node.

Adding both of these cases should make testing of this stuff easier for QE.


odvarkadaniel:
/hold

Have to figure out whether we need to actually use some kind of auth to delete the instances in the cloud providers.

odvarkadaniel:
Test for AWS seems to be deleting the instance on the cloud properly, need to investigate whether the steps after that are correct (find and delete the failed machine and check for rollout).

As we have (AFAIK) only periodics job for AWS, shall I create ones for Azure and GCP? cc @JoelSpeed

odvarkadaniel:
Currently, when we delete an instance on the cloud provider successfully, a node in OpenShift gets into NotReady state and soon followed as non-existent (deleted). Right after that a corresponding machine gets into phase Failed, which is also expected. Right now, we would like to delete the machine as seen [here](https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/c01faca253bec4b53fa1d2bd91d5bd07e065ac7c/test/e2e/periodic_test.go#L75-L78), but we are running into an issue with etcd, which is expecting 3 control plane nodes, but only has two (the third one, new, gets up and running after the new machine gets into Running phase) at the moment, which is preventing this [block of code](https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/c01faca253bec4b53fa1d2bd91d5bd07e065ac7c/test/e2e/periodic_test.go#L32-L34) from deleting the machine as the etcd has not stabilized yet.

odvarkadaniel:
The pre-submit e2e test case with instance being deleted passed for AWS. 
I want to make sure that this case also passes for GCP and Azure.

odvarkadaniel:
Just have checked that this first part of the tests work for AWS, GCP and Azure (delete an instance in the cloud provider and and after we delete the Failed machine, a new one comes up at the index).

test/e2e/framework/framework.go
```diff
@@ -292,6 +328,29 @@ func (f *framework) UpdateDefaultedValueFromCPMS(rawProviderSpec *runtime.RawExt
...
+// DeleteAnInstanceFromCloudProvider deletes an instances from a cloud provider.
+// Currently supported are AWS, Azure and GCP.
+func (f *framework) DeleteAnInstanceFromCloudProvider() error {
+	ctx := f.GetContext()
+	client := f.client
+
+	cpms := &machinev1.ControlPlaneMachineSet{}
+	if err := client.Get(f.GetContext(), f.ControlPlaneMachineSetKey(), cpms); err != nil {
+		return fmt.Errorf("control plane machine set should exist: %w", err)
+	}
```
JoelSpeed, line 340: This is just to check that the CPMS exists, as a safety check?

test/e2e/framework/framework.go
```diff
@@ -770,3 +829,201 @@ func setProviderSpecValue(rawProviderSpec *runtime.RawExtension, value interface
...
+		return "", fmt.Errorf("failed to write credentials to %s: %v", f.Name(), err)
+	}
+
+	return f.Name(), nil
+}
+
+// deleteAWSInstance deletes an instance from the AWS cloud provider.
+func deleteAWSInstance(ctx context.Context, client runtimeclient.Client) error {
+	var credentialsSecret corev1.Secret
+	Expect(client.Get(ctx, runtimeclient.ObjectKey{
```
JoelSpeed, line 871: Framework shouldn't use gomega, just return the error

test/e2e/framework/framework.go
```diff
@@ -770,3 +829,201 @@ func setProviderSpecValue(rawProviderSpec *runtime.RawExtension, value interface
...
+		Namespace: namespaceSecret,
+		Name:      awsCredentialsSecretName,
+	}, &credentialsSecret)).To(Succeed(), "should be able to retrieve the cloud credentials secret for AWS")
+
+	machineSelector := runtimeclient.MatchingLabels(ControlPlaneMachineSetSelectorLabels())
+	machineList := &machinev1beta1.MachineList{}
+	client.List(ctx, machineList, machineSelector)
```
JoelSpeed, line 878: I think we should have a machine passed in rather than selecting ourself

test/e2e/framework/framework.go
```diff
@@ -770,3 +829,201 @@ func setProviderSpecValue(rawProviderSpec *runtime.RawExtension, value interface
...
+}
+
+// deleteAzureInstance deletes an instance from the Azure cloud provider.
+func deleteAzureInstance(ctx context.Context, client runtimeclient.Client) error {
+	machineSelector := runtimeclient.MatchingLabels(ControlPlaneMachineSetSelectorLabels())
+	machineList := &machinev1beta1.MachineList{}
+	client.List(ctx, machineList, machineSelector)
+
+	var credentialsSecret corev1.Secret
+	Expect(client.Get(ctx, runtimeclient.ObjectKey{
```
JoelSpeed, line 931: Shouldn't use gomega here

test/e2e/framework/framework.go
```diff
@@ -770,3 +829,201 @@ func setProviderSpecValue(rawProviderSpec *runtime.RawExtension, value interface
...
+
+	clientSecret, ok := credentialsSecret.Data["azure_client_secret"]
+	if !ok {
+		return fmt.Errorf("could not get clientSecret from Azure credentials secret")
+	}
+
+	tenantID, ok := credentialsSecret.Data["azure_tenant_id"]
+	if !ok {
+		return fmt.Errorf("could not get tenantID from Azure credentials secret")
+	}
```
JoelSpeed, line 959: Perhaps move all of the client setup to a helper function?

test/e2e/framework/framework.go
```diff
@@ -770,3 +829,201 @@ func setProviderSpecValue(rawProviderSpec *runtime.RawExtension, value interface
...
+		return fmt.Errorf("should be able to delete an Azure instance: %w", err)
+	}
+
+	return nil
+}
+
+// deleteGCPInstance deletes an instance from the GCP cloud provider.
+func deleteGCPInstance(ctx context.Context, client runtimeclient.Client, logger logr.Logger) error {
+	var credentialsSecret corev1.Secret
+	Expect(client.Get(ctx, runtimeclient.ObjectKey{
```
JoelSpeed, line 984: Shouldn't use gomega here

test/e2e/helpers/cases.go
```diff
@@ -228,6 +228,51 @@ func ItShouldOnDeleteReplaceTheOutDatedMachineWhenDeleted(testFramework framewor
...
+
+		machine, err := machineForIndex(testFramework, index)
+		Expect(err).ToNot(HaveOccurred(), "control plane machine should exist")
+
+		By(fmt.Sprintf("The machine at index %d should be deleting phase", index))
+		Eventually(komega.Object(machine), 10*time.Second).Should(
+			HaveField("Status.Phase", HaveValue(Equal("Deleting"))),
+		)
+
+		// Make sure the CPMS exists before we delete the Machine, just in case.
```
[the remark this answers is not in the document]
odvarkadaniel: Yeah, going to remove that comment/rewrite.

test/e2e/helpers/cases.go
```diff
@@ -228,6 +228,51 @@ func ItShouldOnDeleteReplaceTheOutDatedMachineWhenDeleted(testFramework framewor
...
+		cpms := &machinev1.ControlPlaneMachineSet{}
+		Expect(k8sClient.Get(ctx, testFramework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
+
+		// Deleting the Machine triggers a rollout, give the rollout 30 minutes to complete.
```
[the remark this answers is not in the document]
odvarkadaniel: Yes.

test/e2e/periodic_test.go
```diff
@@ -45,5 +48,26 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+			BeforeEach(func() {
+				client := testFramework.GetClient()
+				machineList := &machinev1beta1.MachineList{}
+				machineSelector := runtimeclient.MatchingLabels(framework.ControlPlaneMachineSetSelectorLabels())
+
+				By("Getting a list of all control plane machines")
+				Expect(client.List(testFramework.GetContext(), machineList, machineSelector)).To(Succeed(), "should be able to retrieve list of control plane machines")
+
+				By("Deleting an instance from the cloud provider")
+				Expect(testFramework.DeleteAnInstanceFromCloudProvider()).To(Succeed())
```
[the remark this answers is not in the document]
odvarkadaniel: Makes sense, will change.

test/e2e/framework/framework.go
odvarkadaniel: Makes sense.

test/e2e/framework/framework.go
odvarkadaniel: +1

test/e2e/framework/framework.go
odvarkadaniel: Yes.

odvarkadaniel:
Had to bump golangci-lint package to fix the CI linting issues. The linter is becoming more and more invasive and checks really unnecessary things. I think we might want to take a look into configuring it better to stop targeting things like putting backslash as a constant, cc @JoelSpeed.

odvarkadaniel:
> /hold
> 
> I don't know what's up but this is spamming my inbox

For some reason, `gotoolchain v1.x.x` appeared in the `go.mod`

odvarkadaniel:
> > /hold
> > I don't know what's up but this is spamming my inbox
> 
> For some reason, `gotoolchain v1.x.x` appeared in the `go.mod`

CI should be fixed now, lets see.

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,76 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+			BeforeEach(func() {
+				client := testFramework.GetClient()
+				machineList := &machinev1beta1.MachineList{}
+				machineSelector := runtimeclient.MatchingLabels(framework.ControlPlaneMachineSetSelectorLabels())
+
+				By("Getting a list of all control plane machines")
+				Expect(client.List(testFramework.GetContext(), machineList, machineSelector)).To(Succeed(), "should be able to retrieve list of control plane machines")
+
+				By("Deleting an instance from the cloud provider")
+				Expect(testFramework.DeleteAnInstanceFromCloudProvider(&machineList.Items[0])).To(Succeed())
```
JoelSpeed, line 65: Can you take a variable rather than `&machineList.Items[0]` several times? Also, I'd expect a test to check that the machine is actually index 0, or you need to deliberately pick index 0 from the list

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,76 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				By("Waiting for the machine to get into Failed phase")
+				Eventually(komega.Object(&machineList.Items[0]), 10*time.Minute).Should(HaveField("Status.Phase", HaveValue(Equal("Failed"))))
+
+				By("Deleting a control plane machine in phase Failed at index 0")
+				Expect(client.Delete(testFramework.GetContext(), &machineList.Items[0])).To(Succeed())
+			})
+
+			helpers.ItShouldReplaceTheOutDatedMachineInDeleting(testFramework, 0)
```
JoelSpeed, line 74: Not obvious why this is index 0, should make sure the machine you are operating is index 0

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,76 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+				By("Waiting for the node to get into NotReady phase")
+				Eventually(komega.Object(node), 10*time.Minute).Should(
+					WithTransform(func(node *corev1.Node) []corev1.ConditionStatus {
+						statuses := []corev1.ConditionStatus{}
+						for _, condition := range node.Status.Conditions {
+							statuses = append(statuses, condition.Status)
+						}
+
+						return statuses
+					}, ContainElements(corev1.ConditionUnknown, corev1.ConditionFalse)))
```
JoelSpeed, line 111: What exactly are you checking here? To fit with the rest of the tests here, we would just use gomega constructs and be explicit about which conditions should have which statuses
```suggestion
				Eventually(komega.Object(node), 10*time.Minute).Should(HaveField("Status.Conditions", 
					ContainElements(
						SatisfyAll(HaveField("Type", Equal(<something>),HaveField("Status", Equal(corev1.ConditionUnknown))),
						SatisfyAll(HaveField("Type", Equal(<something>),HaveField("Status", Equal(corev1.ConditionFalse))),
					),
				))
```

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,76 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				By("Deleting node's control plane machine")
```
JoelSpeed, line 113: Nit
```suggestion
				By("Deleting the node's control plane machine")
```

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,76 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				By("Deleting a control plane machine in phase Failed at index 0")
+				Expect(client.Delete(testFramework.GetContext(), &machineList.Items[0])).To(Succeed())
+			})
+
+			helpers.ItShouldReplaceTheOutDatedMachineInDeleting(testFramework, 0)
+		})
+
+		Context("and a node with terminated kubelet", func() {
+			var delObjects map[string]runtimeclient.Object
```
JoelSpeed, line 78: I believe there's a global cleanup routine in an after each that takes a list of object types to clean up, better to re-use that please

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/cluster-control-plane-machine-set-operator
 
-go 1.18
+go 1.21
```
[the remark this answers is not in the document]
odvarkadaniel: This was causing the CI issue along with the `toolchain go1.21.1` - it was for some reason introduced by bumping golangci-lint package. This is already reverted and the go version is currently 1.20.

go.mod
```diff
@@ -1,6 +1,8 @@
...
+
+toolchain go1.21.1
```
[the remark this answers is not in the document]
odvarkadaniel: This was causing the CI issue along with the go version being 1.21 - it was for some reason introduced by bumping golangci-lint package. This is already reverted and the go version is currently 1.20.

test/e2e/periodic_test.go
odvarkadaniel: With the state of the cluster at the time of this test running, I expect it to be in index 0.
Introducing a test to check that is a good idea, will do that, just in case.

test/e2e/periodic_test.go
odvarkadaniel: So plugging in a machine (that I made sure that is index 0 machine) instead of the index is better?

JoelSpeed: You plug in index 0 directly, as long as you are certain that the machine you operated on was index 0, but right now the before each doesn't assert that anywhere

test/e2e/periodic_test.go
odvarkadaniel: Is this actually correct?

We have `nodeObj.Status.Conditions` as a list of `NodeCondition` and we need to check that every `NodeCondition.Status` is either `Unknown` or `False`. What didn't you like about my solution? Not idiomatic? Thanks for clarifying!

JoelSpeed: I wasn't actually sure what you were trying to achieve with that check, but, if you wanted to check that each condition was Unknown or False, your current approach also won't work, since it would pass for `{Unknown,False,True,True}`

If you want to make sure each one is either Unknown or False, then you need
```suggestion
				Eventually(komega.Object(node), 10*time.Minute).Should(HaveField("Status.Conditions", 
					HaveEach(HaveField("Status",
						SatisfyAny(Equal(corev1.ConditionUnknown), Equal(corev1.ConditionFalse)),
					)),
				))
```
This checks that for every element in the list, the Status field is Unknown or False

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,95 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				By("Getting a list of all control plane machines")
+				Expect(client.List(testFramework.GetContext(), machineList, machineSelector)).To(Succeed(), "should be able to retrieve list of control plane machines")
+
+				machine = &machineList.Items[0]
+
+				By("Checking that the machine is actually in index 0")
+				machineIdx, err := helpers.MachineIndex(*machine)
+				Expect(machineIdx).To(Equal(0))
+				Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 72: Switch these please, errors are always checked first else you risk masking the error with the failure of the assertion to equal 0

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,95 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				By("Getting a list of all control plane machines")
+				Expect(client.List(ctx, machineList, machineSelector)).To(Succeed(), "should be able to retrieve list of control plane machines")
+
+				machine = &machineList.Items[0]
+
+				By("Checking that the machine is actually in index 0")
+				machineIdx, err := helpers.MachineIndex(*machine)
+				Expect(machineIdx).To(Equal(0))
+				Expect(err).ToNot(HaveOccurred())
```
JoelSpeed, line 112: Switch as above

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,95 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				idx = machineIdx
+
+				By("Getting the node from the machine")
+				Expect(client.Get(ctx, types.NamespacedName{
+					Namespace: framework.MachineAPINamespace,
```
JoelSpeed, line 118: Nodes are not namespaced

test/e2e/framework/framework.go
```diff
@@ -292,6 +328,195 @@ func (f *framework) UpdateDefaultedValueFromCPMS(rawProviderSpec *runtime.RawExt
...
+}
+
+// TerminateKubelet terminates kubelet on a node that is being referenced by the input machine.
+func (f *framework) TerminateKubelet(node *corev1.Node, delObjects map[string]runtimeclient.Object) error {
+	client := f.GetClient()
+	ctx := f.GetContext()
+
+	serviceAccount, err := createServiceAccount(ctx, client)
+	if err != nil {
+		return err
```
JoelSpeed, line 361: Please wrap all the errors here else we won't be able to identify which one was causing issues

odvarkadaniel: The function `createServiceAccount` already return the wrapped err, do we need to do that twice here?

JoelSpeed: In general it's good practice to wrap every error, that way if something changes elsewhere, you still have the trace as you climb the stack

JoelSpeed:
Changes look good, we are blocked currently on a different bug though right?

odvarkadaniel:
> Changes look good, we are blocked currently on a different bug though right?

Yes, this is blocked on an etcd related bug.

odvarkadaniel:
The etcd bug got merged last week.

/retest

test/e2e/periodic_test.go
```diff
@@ -45,5 +52,94 @@ var _ = Describe("ControlPlaneMachineSet Operator", framework.Periodic(), func()
...
+
+				By("Getting a list of all control plane machines")
+				Expect(client.List(ctx, machineList, machineSelector)).To(Succeed(), "should be able to retrieve list of control plane machines")
+
+				machine = &machineList.Items[0]
+
+				By("Checking that the machine is actually in index 0")
+				machineIdx, err := helpers.MachineIndex(*machine)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(machineIdx).To(Equal(0))
```
JoelSpeed, line 112: Please add a note here to explain the expectation
```suggestion
				Expect(machineIdx).To(Equal(0), "Expected first machine to have index 0")
```
I expect this is fragile and is why the tests are failing right now, I think you may want a better way to extract the index 0 machine from this list and in the `BeforeEach`

odvarkadaniel: I suspect that the instance type increase test that runs before this is non-deterministic (we don't know what machine gets updated first time-wise).

Do we even need to check the idx? We always get the machine at 0 from the machineList. I don't really get what machine index this actually is (we get a correct idx from the `helpers.MachineIndex` to plug into `	helpers.ItShouldReplaceTheOutDatedMachineInDeleting` later). What functionality are we even testing by this check?

JoelSpeed: > We always get the machine at 0 from the machineList

I don't think that's correct looking at the periodic-pre. I think this is failing because you've got index 2 from the list rather than index 0. I suspect, because there are changes to the cluster, that index 0 isn't alphabetically first always, and so this is causing a flake.

You need to make sure you're working within Index 0 here as this is the one you've told to be deleted, so, if for any reason the list order is not with the index 0 first, you'll fail this test later in weird ways.

This is definitely an important check IMO

odvarkadaniel: Came up with something. I will see how CI behaves.

Loop through the control plane machines until we find the expected index and return this machine and work with it.

odvarkadaniel:
/retest

CI is slow.

JoelSpeed (commented):
Something tells me there's an actual issue with the unit tests here that we need to look to resolve

odvarkadaniel:
> Something tells me there's an actual issue with the unit tests here that we need to look to resolve

Unit tests pass locally on my machine and I also don't see anything in the logs, that would suggest what is going wrong.

JoelSpeed:
Have you tried executing the unit tests from with a container? As we do in CI? That would isolate your own environment and maybe be more representative?

Seems it's consistently failing at starting the kube-apiserver binary right? You could try adding config to envtest to have it spit out the logs from the kube-apiserver binary, perhaps there's an issue?

odvarkadaniel:
CI passed.

/unhold

PR #242, opened 2023-09-18
OCPCLOUD-2169: Add events for machine creation/deletion

odvarkadaniel:
This PR introduces simple wrapper over client-go's events package to make it easier for us to emit events whenever needed throughout the controller.


odvarkadaniel:
Events for machine creation/deletion, https://github.com/openshift/origin/pull/28254.

cc/ @JoelSpeed

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -544,6 +545,8 @@ func (m *openshiftMachineProvider) CreateMachine(ctx context.Context, logger log
 		"failureDomain", providerConfig.ExtractFailureDomain().String(),
 	)
 
+	recorder.Eventf(machine, "MachineCreatedSuccessfully", "created machine: %s", machine.Name)
```
JoelSpeed, line 548: I think we should emit the event on the CPMS, since this is the object taking the action. Then when you `oc describe controlplanemachineset` it will have all of the created and deleted events together right?

Worth checking though, what events do a Deployment or StatefulSet create when they create a pod? Are they on the pods or on the parent?

pkg/recorder/recorder.go
```diff
@@ -0,0 +1,78 @@
...
+	//nolint: gochecknoglobals
+	defaultRecorder cgrecord.EventRecorder
+)
+
+func init() {
+	defaultRecorder = new(cgrecord.FakeRecorder)
+}
+
+// InitFromRecorder initializes the global default recorder. It can only be called once.
+func InitFromRecorder(recorder cgrecord.EventRecorder) {
```
JoelSpeed, line 46: Does controller runtime not provide an event recorder wrapper already? Or is this just convenience on top?

odvarkadaniel: It does, but this makes sure for the recorder to be singleton. I guess we don't want more than 1 recorders to be initiated, so this is just convenient to have IMO. Wdyt?

JoelSpeed: Typically I would make the recorder a member of the reconciler struct, that's what we've done in other places, and in that pattern you don't need to worry about this sync once stuff. Instead you will need to update tests to ensure a fake recorder is used though

odvarkadaniel: Yes, I have seen, but I would argue that this might be an easier way to do that and is IMO easier to set up.

pkg/machineproviders/providers/machineproviders_test.go
```diff
@@ -34,6 +35,8 @@ import (
 )
 
 var _ = Describe("MachineProviders", func() {
+	recorder := record.NewFakeRecorder(10)
```
JoelSpeed, line 38: Whats the 10 for?

This probably needs to be a var at this scope, but with a `BeforeEach` to initialise it for every test run to a fresh recorder

Is it possible to check the output of the test events? Would be good to see, when we expect the Machine to be deleted or created, that the expected event is emitted

JoelSpeed: > I can create a util function that check what comes into the event channel and then just check whether the right event came in. That would just return true/false or an error.

What output does the fake recorder give you?

JoelSpeed: The fake recorder stores events as a channel, so you should be able to use ginkgo to match and event from that channel

odvarkadaniel:
/hold

Event is not appearing in the cpms after the changes.

cmd/control-plane-machine-set-operator/main.go
```diff
@@ -128,6 +128,9 @@ func main() { //nolint:funlen,cyclop
 		os.Exit(1)
 	}
 
+	// Initialize event recorder.
+	// recorder.InitFromRecorder(mgr.GetEventRecorderFor("control-plane-machine-set-controller"))
```
JoelSpeed, line 132: Drop before merge

cmd/control-plane-machine-set-operator/main.go
```diff
@@ -146,6 +149,7 @@ func main() { //nolint:funlen,cyclop
 		Namespace:      managedNamespace,
 		OperatorName:   "control-plane-machine-set",
 		ReleaseVersion: getReleaseVersion(setupLog),
+		Recorder:       mgr.GetEventRecorderFor("controlplanemachineset-controller"),
```
JoelSpeed, line 152: We hyphenate in other places don't we?
```suggestion
		Recorder:       mgr.GetEventRecorderFor("control-plane-machine-set-controller"),
```

cmd/control-plane-machine-set-operator/main.go
```diff
@@ -146,6 +146,7 @@ func main() { //nolint:funlen,cyclop
 		Namespace:      managedNamespace,
 		OperatorName:   "control-plane-machine-set",
 		ReleaseVersion: getReleaseVersion(setupLog),
+		Recorder:       mgr.GetEventRecorderFor("control-plane-machine-set-controller-controller"),
```
damdo, line 149: We have 2 controllers in the CPMSO.
Could we stick to some other name?

pkg/controllers/controlplanemachineset/controller.go
```diff
@@ -144,6 +146,7 @@ func (r *ControlPlaneMachineSetReconciler) SetupWithManager(mgr ctrl.Manager) er
 
 	// Set up API helpers from the manager.
 	r.Scheme = mgr.GetScheme()
+	r.Recorder = mgr.GetEventRecorderFor("controlplanemachineset-controller")
```
damdo, line 149: Same

cmd/control-plane-machine-set-operator/main.go
JoelSpeed: The generator currently isn't emitting events so I would expect this to be `control-plane-machine-set-controller` and the other to be `control-plane-machine-set-generator` right?

damdo: Yes, that sounds better 👍

JoelSpeed, line 149: You appear to be setting the Recorder twice, once here and once in SetupWithManager, we only need one

odvarkadaniel: Overlooked that. But same thing is happening for `.Scheme`, right? I would rather initialise that in the main.go file instead of `SetupWithManager`, wdyt?

JoelSpeed: Yeah it is, I suspect that's so that tests have it set up, does it break if you drop it from the SetupWithManager? 

Make sure to update the name in main

pkg/machineproviders/providers/openshift/machine/v1beta1/provider.go
```diff
@@ -544,6 +549,8 @@ func (m *openshiftMachineProvider) CreateMachine(ctx context.Context, logger log
 		"failureDomain", providerConfig.ExtractFailureDomain().String(),
 	)
 
+	m.recorder.Eventf(cpms, corev1.EventTypeNormal, "Created", "created machine: %s", machine.Name)
```
JoelSpeed, line 552: Why not `CreatedMachine`? Might be a bit more helpful? How does this present currently?

odvarkadaniel: I wanted to make this like other events that are being emitted by other controllers. They usually use just verbs such as "Created", "Deleted", "Updated", ...

JoelSpeed: Ok, that works for me then, thanks for checking on those other controllers, +1 for consistency

PR #243, opened 2023-09-21
OCPCLOUD-2169: Check CPMS events are not present when we expect error

odvarkadaniel:
Follow up to #242.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
@@ -1881,6 +1881,10 @@ var _ = Describe("MachineProvider", func() {
 					Consistently(komega.Get(machine)).Should(Succeed())
 				})
 
+				It("should not receive an event", func() {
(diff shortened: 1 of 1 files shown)


pkg/machineproviders/providers/openshift/machine/v1beta1/provider_test.go
```diff
@@ -1881,6 +1881,10 @@ var _ = Describe("MachineProvider", func() {
 					Consistently(komega.Get(machine)).Should(Succeed())
 				})
 
+				It("should not receive an event", func() {
+					Expect(recorder.Events).Should(Not(Receive(ContainSubstring("Deleted"))))
```
JoelSpeed, line 1885: We expect no event at all right? So you can just say it shouldn't receive 
```suggestion
					Expect(recorder.Events).Should(Not(Receive()))
```

odvarkadaniel: Addressed

PR #244, opened 2023-10-04
OCPBUGS-20070: fix: e2e: add gcp custom type to test framework

damdo:
Fixes a test framework failure by adding the GCP instance custom type support to it.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -714,7 +714,7 @@ func increaseOpenStackInstanceSize(rawProviderSpec *runtime.RawExtension, provid
 //nolint:cyclop
 func nextGCPMachineSize(current string) (string, error) {
 	// Regex to match the GCP machine size string.
-	re := regexp.MustCompile(`^(?P<family>[0-9a-z]+)-(?P<subfamily>[0-9a-z]+(-(?P<subfamilyflavor>[a-z]+))?)-(?P<multiplier>[0-9]+)(-(?P<multiplier2>[0-9]+))?`)
+	re := regexp.MustCompile(`^(?P<family>[0-9a-z]+)-(?P<subfamily>[0-9a-z]+(-(?P<subfamilyflavor>[a-z]+))?)-(?P<multiplier>[0-9\.]+)(-(?P<multiplier2>[0-9]+))?`)
 
 	subexpNames := re.SubexpNames()
 	values := re.FindStringSubmatch(current)
@@ -735,20 +735,23 @@ func nextGCPMachineSize(current string) (string, error) {
 	family, okFamily := result["family"]
 	subfamily, okSubfamily := result["subfamily"]
 	_, okMultiplier := result["multiplier"]
+	subfamilyflavor, okSubfamilyflavor := result["subfamilyflavor"]
 
-	if !(okFamily && okSubfamily && okMultiplier) {
+	if !(okFamily && okSubfamily && okMultiplier && okSubfamilyflavor) {
 		return "", fmt.Errorf("%w: %s", errInstanceTypeUnsupportedFormat, current)
 	}
 
-	multiplier, err := strconv.Atoi(result["multiplier"])
+	multiplier, err := strconv.ParseFloat(result["multiplier"], 64)
 	if err != nil {
 		// This is a panic because the multiplier should always be a number.
-		panic("failed to convert multiplier to int")
+		panic("failed to convert multiplier to float")
 	}
 
 	var multiplier2 int
 
 	if val, okMultiplier2 := result["multiplier2"]; okMultiplier2 && val != "" {
+		var err error
+
 		multiplier2, err = strconv.Atoi(val)
 		if err != nil {
 			// This is a panic because the multiplier2 should always be a number.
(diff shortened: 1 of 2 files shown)


damdo:
@RadekManak the only periodic pre available is for AWS.

damdo:
But it looks like in the presubmits we do run the increase instance size step.
https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/a7e0190f60a8777430752e68604e18232b38c76e/test/e2e/periodic_test.go#L40
So let's see if that passes, and if we get those `e2-custom` machines for that.
Otherwise we might need to try and test it in a different way.

test/e2e/framework/framework.go
```diff
@@ -708,32 +709,61 @@ func increaseOpenStackInstanceSize(rawProviderSpec *runtime.RawExtension, provid
...
 
 // nextGCPVMSize returns the next GCP machine size in the series.
-// The Machine sizes being used are in format <e2|n2|n1>-standard-<number>.
+// The Machine sizes being used are in format <e2|n2|n1>-<subfamily(-subfamilyflavor(optional))>-<number>(-<number>(optional)).
+//
+//nolint:cyclop
 func nextGCPMachineSize(current string) (string, error) {
 	// Regex to match the GCP machine size string.
-	re := regexp.MustCompile(`(?P<family>[0-9a-z]+)-standard-(?P<multiplier>[0-9]+)`)
+	re := regexp.MustCompile(`^(?P<family>[0-9a-z]+)-(?P<subfamily>[0-9a-z]+(-(?P<subfamilyflavor>[a-z]+))?)-(?P<multiplier>[0-9]+)(-(?P<multiplier2>[0-9]+))?`)
```
odvarkadaniel, line 717: Let's have an input string `current = e2-custom-6-16384`, then this regular expression `re` groups the values into following groups:
- family => e2
- subfamily => custom
- multiplier => 6
- '6' => -16384
- multiplier => 16384

If I see this correctly, then the check `if len(values) != len(subexpNames)` will result in an error.

values = `[family subfamily  subfamilyflavor multiplier  multiplier2]`
subexpNames = ` [e2-custom-6-16384 e2 custom 6 -16384 16384]`
values.len() = 5
subexpNames = 6

Am I missing something here?

test/e2e/framework/framework.go
```diff
@@ -708,32 +709,61 @@ func increaseOpenStackInstanceSize(rawProviderSpec *runtime.RawExtension, provid
...
 
+	subexpNames := re.SubexpNames()
 	values := re.FindStringSubmatch(current)
-	if len(values) != 3 {
+	result := make(map[string]string)
+
+	// The number of named regex subexpressions must always match the number of submatches.
+	if len(values) != len(subexpNames) {
```
odvarkadaniel, line 724: This is the check that is concerning me.

test/e2e/framework/framework.go
damdo: It is true that "on-paper" we have 5 named subexpressions and not 6, but  there is a catch [here](https://pkg.go.dev/regexp#Regexp.SubexpNames):

> SubexpNames returns the names of the parenthesized subexpressions in this Regexp. The name for the first sub-expression is names[1], so that if m is a match slice, the name for m[i] is SubexpNames()[i]. Since the Regexp as a whole cannot be named, names[0] is always the empty string. The slice should not be modified. 

Which means the `len(subexpNames)` is always going to be _number-of-defined-names_+1, hence that should always match the number of submatches.

test/e2e/framework/framework.go
damdo: See https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/244#discussion_r1346156637

test/e2e/framework/framework.go
odvarkadaniel: Thanks for clarification, totally missed that part of the docs. Makes sense to me.

test/e2e/framework/framework.go
```diff
@@ -749,6 +779,12 @@ func setNextGCPMachineSize(current, family string, multiplier int) (string, erro
 		multiplier = 128
 	case multiplier >= 128:
 		return "", fmt.Errorf("%w: %s", errInstanceTypeNotSupported, current)
+	case strings.HasPrefix(subfamily, "custom"):
+		if multiplier == 0 || multiplier2 == 0 {
+			return "", fmt.Errorf("%w: %s", errInstanceTypeNotSupported, current)
+		}
+
+		return fmt.Sprintf("%s-%s-%d-%d", family, subfamily, multiplier+1, multiplier2+256), nil
```
JoelSpeed, line 787: Where did the +1 and +256 come from?

odvarkadaniel: I believe it comes from this doc for custom GCP instances: https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type

JoelSpeed: Reading that, it suggests that memory is relatively free in what we can spec, so +256 is ok, but it suggests that CPUs can be "1 to 96 or any **even** value in between".

I don't think we are accounting for the even part of the above statement

odvarkadaniel: According to: https://cloud.google.com/compute/docs/general-purpose-machines#custom_machine_types

> E2 custom machine types support predefined platforms with Intel or AMD EPYC processors. You can create E2 custom machine types with vCPUs in multiples of 2, up to 32 vCPUs. The minimum acceptable number of vCPUs for a VM is 2.

There is also a constraint for memory - cpu cores ratio (shown in this example of invalid instance):

> 32 vCPUs, 1 GB of total memory. Invalid because the ratio of vCPUs to memory is incorrect. The acceptable ratio is 0.5 GB of memory to 1 vCPU.

damdo: Ahh good catch @JoelSpeed. I looked at the [limitations section](https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#limitations) but couldn't find anything regarding CPU.
Thanks for flagging, will update this.

damdo: I've updated it by taking into account **all** the limitations described in the doc that @odvarkadaniel shared.

PR #245, opened 2023-10-09
Set stabilisationTimeout to 60 In rolling update

itzikb-redhat:
It may take up to 60 minutes for operators to be up in slow setups

### DIFF
--- a/test/e2e/helpers/cases.go
+++ b/test/e2e/helpers/cases.go
@@ -102,7 +102,7 @@ func ItShouldPerformARollingUpdate(opts *RollingUpdatePeriodicTestOptions) {
 		By("Control plane machine replacement completed successfully")
 
 		By("Waiting for the cluster to stabilise after the rollout")
-		stabilisationTimeout := 30 * time.Minute
+		stabilisationTimeout := 60 * time.Minute
 		if opts.StabilisationTimeout.Seconds() != 0 {
 			stabilisationTimeout = opts.StabilisationTimeout
 		}


JoelSpeed:
Can you define a slow setup? What platform are you running? Which operators are taking time to rollout? Do you have failure messages from your tests that lead you to increasing this timeout?

itzikb-redhat:
> Can you define a slow setup? What platform are you running? Which operators are taking time to rollout? Do you have failure messages from your tests that lead you to increasing this timeout?

It's Openshift on Openstack. I'll rerun and add the details

itzikb-redhat:
@JoelSpeed [cluster-control-plane-machine-set-operator_e2e-periodic.log](https://github.com/openshift/cluster-control-plane-machine-set-operator/files/12877837/cluster-control-plane-machine-set-operator_e2e-periodic.log)

JoelSpeed:
@itzikb-redhat Reviewing the log that you've posted, this timeout isn't what is causing the issue here. There's seems to be an issue with the rollout of the third index, as it was told to "stop trying" after 0.146 seconds. That indicates to me that either an earlier context timed out (possibly) or there was an issue with the CPMS.

Please dig into the error and see if you can work out the timings and understand which timeout actually failed here/why the check failed

/hold

itzikb-redhat:
@JoelSpeed  Can you please take a look at [cluster-control-plane-machine-set-operator_e2e-periodic.log](https://github.com/openshift/cluster-control-plane-machine-set-operator/files/13828622/cluster-control-plane-machine-set-operator_e2e-periodic.log)

JoelSpeed:
The latest failure was because the operators were not stable from a previous test/from cluster init, it failed on https://github.com/openshift/cluster-control-plane-machine-set-operator/blob/ad039c1e94e308a27af01adfef81af9ff566bde5/test/e2e/periodic_test.go#L30C11-L30C52 which is where the 10 minute timeout comes from

JoelSpeed:
Looking at the log in fact, it looks like the cluster wasn't stable when the test suite started, which is odd since the cluster shouldn't complete bootstrap until the cluster is stable, at which point the tests should then start. Is it common that the operators make changes right at the beginning of the cluster lifecycle?

itzikb-redhat:
@JoelSpeed I'm not sure , from what I looked at in the log, the operators were ok at the start of the test.

PR #257, opened 2023-10-20
azure: update Zone to be a string and not a pointer anymore

EmilienM:
- [vendoring] bump cluster-api-actuator-pkg/testutils to latest
- azure: Zone is now a String (not a pointer)


EmilienM:
done via https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/234

PR #269, opened 2023-12-14
OCPCLOUD-1894: Implement admission warning for missing target pools

RadekManak:
This PR implements admission warning for when CPMS providerSpec does not have targetPools field set. This needs to be warning instead of error because on private clusters we don't use target pools.

The warnings are only available as log messages in custom KubeAPIWarningLogger logger that is instantiated from parent logger for controller-runtime when creating client. I added a hook to the parent logger to count the warning messages to enable testing.


JoelSpeed (commented):
/lgtm
/approve

PR #278, opened 2024-02-09
OCPBUGS-29249: Never delete a Machine when there's a single Machine in an index

JoelSpeed:
When applying a MachineConfig update, and a CPMS update at the same time, it is possible that the CPMS will prematurely delete a later index, when it is not yet ready to be replaced.

Applying a MachineConfig will drain and reboot the machine, for some period of time, this means the node goes unready.
If we also apply a CPMS change, then all indexes suddenly become outdated.

While replacing index 0, I observed that MCO was updating index 1, therefore the outdated index 1, suddenly became outdated and non ready. Our deletion logic decides that this machine is fit for removal, however, because it has no replacement, this is bad for the cluster. The etcd operator protection does step in, so in my experience, the cluster eventually recovered, but at points, we were down to 2 functional indexes which is not ideal.

To prevent ever getting into a situation like this, we should never delete a machine if an index has fewer than 2 machines. There is no scenario where we would ever want to delete the last machine in an index, unless and until we horiztonal vertical scaling down of the control plane. This is not presently on the roadmap, so this fix should be appropriate for a while.

### DIFF
--- a/pkg/controllers/controlplanemachineset/updates.go
+++ b/pkg/controllers/controlplanemachineset/updates.go
@@ -346,6 +346,11 @@ func (r *ControlPlaneMachineSetReconciler) waitForRemoveMachine(logger logr.Logg
 }
 
 func (r *ControlPlaneMachineSetReconciler) deleteReplacedMachines(ctx context.Context, logger logr.Logger, machineProvider machineproviders.MachineProvider, machines []machineproviders.MachineInfo) (bool, ctrl.Result, error) {
+	if len(machines) < 2 {
+		// No need to delete any machines if there are less than 2 machines in an index.
+		return false, ctrl.Result{}, nil
+	}
+
 	machinesNeedingReplacement := needReplacementMachines(machines)
 	machinesUpdated := updatedMachines(machines)
 	machinesOutdatedNonReady := nonReadyMachines(machinesNeedingReplacement)
--- a/pkg/controllers/controlplanemachineset/updates_test.go
+++ b/pkg/controllers/controlplanemachineset/updates_test.go
@@ -397,6 +397,46 @@ var _ = Describe("reconcileMachineUpdates", func() {
 					}
 				},
 			}),
+			Entry("with updates required in a single index, and the outdated index is not ready", rollingUpdateTableInput{
+				cpmsBuilder: cpmsBuilder.WithReplicas(3),
+				machineInfos: map[int32][]machineproviders.MachineInfo{
+					0: {updatedMachineBuilder.WithIndex(0).WithMachineName("machine-0").WithNodeName("node-0").Build()},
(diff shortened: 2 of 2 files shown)


pkg/controllers/controlplanemachineset/updates.go
```diff
@@ -346,6 +346,11 @@ func (r *ControlPlaneMachineSetReconciler) waitForRemoveMachine(logger logr.Logg
 }
 
 func (r *ControlPlaneMachineSetReconciler) deleteReplacedMachines(ctx context.Context, logger logr.Logger, machineProvider machineproviders.MachineProvider, machines []machineproviders.MachineInfo) (bool, ctrl.Result, error) {
+	if len(machines) < 2 {
```
mdbooth, line 349: Is there any way to obtain the current minimum quorum size rather than hard-coding 2 here?

Either way, this is clearly safer.

JoelSpeed: It shouldn't matter I don't think. CPMS works in terms of indexes, it has no care for etcd. It's designed to have either 3 or 5 indexes depending on the number of control plane instances, only 3 being supported. Each index, in a happy and healthy running state, should only ever have 1 machine in it. If there's more than 1 machine, either, we are going through an update, or someone accidentally created another machine in the index somehow.

If there's less than 1, that means the cluster is missing an index, you'll see this with `unavaiableReplicas`. In a healthy cluster that value should always be 0, so we should always have at least 1 machine in an index.

I'm pretty happy given the indexing grouping that this is appropriate, and that we don't need to worry about the quorum.

In a 5 control plane cluster, we would have 5 indexes, again though, each index would be 1 machine ideally.

pkg/controllers/controlplanemachineset/updates_test.go
```diff
@@ -516,6 +516,60 @@ var _ = Describe("reconcileMachineUpdates", func() {
 				},
 				expectedResult: ctrl.Result{RequeueAfter: 5 * time.Second},
 			}),
+			Entry("with updates are required in multiple indexes, but the replacement machine is pending, and a later index is not ready", rollingUpdateTableInput{
```
damdo, line 519: Nice, this one matches the case where Matt found himself into and your reproducer.
I'd add also one base case to mark the edge of the minimum viable "synthetic" reproducer.

a) `with updates are required in a single index, but there is only one machine and it is not ready` -> should not delete (the sole outdated and unready machine as it should create the replacement first (respecting the surge))

pkg/controllers/controlplanemachineset/updates.go
damdo: +1 to what @JoelSpeed said

damdo (commented):
/approve
With one suggestion, non blocking.

damdo (commented):
/lgtm

PR #287, opened 2024-04-09
OCPBUGS-31808: Fix ExportFailureDomain to handle empty platform spec

vr4manta:
[OCPBUGS-31808](https://issues.redhat.com/browse/OCPBUGS-31808)

### Changes
- Added logic to handler older infrastructure definitions where platform spec is not present.

### DIFF
--- a/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/vsphere.go
+++ b/pkg/machineproviders/providers/openshift/machine/v1beta1/providerconfig/vsphere.go
@@ -163,6 +163,11 @@ func (v VSphereProviderConfig) ExtractFailureDomain() machinev1.VSphereFailureDo
 		return machinev1.VSphereFailureDomain{}
 	}
 
+	// Older OCP installs will not have PlatformSpec set for infrastructure.
+	if v.infrastructure.Spec.PlatformSpec.VSphere == nil {
+		return machinev1.VSphereFailureDomain{}
+	}
+
 	failureDomains := v.infrastructure.Spec.PlatformSpec.VSphere.FailureDomains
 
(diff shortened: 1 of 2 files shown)


damdo:
@rvanderp3 @vr4manta thanks for putting this together!
/test unit

damdo:
I tested units on main and they work.
Not sure whether the issues we are seeing here are flakes or actual issues.
Let's try one more time, otherwise we'll need to look into why they are broken
/test unit

vr4manta:
@damdo the unit tests have been flaky for a while.  I have cleaned up the ones related to vsphere, but there are others that seem to randomly fail from how they were developed.  Currently the unit test is green.  I'm rerunning the other reds to just see if I can get them green.

vr4manta:
the error in e2e-vsphere-operator test looks like it may be related to other issue customer was hitting.  I've pinged out in channel for team to see if there is anyone already looking at this issue.  I am also curious if fixing this issue now exposes the error seen.  For upgrade scenario, the current fix solves that.  Creating / modifying existing CPMS seems to generate the template format error.

damdo (commented):
/approve

/hold
until vsphere E2Es are green

vr4manta:
@damdo I have a change coming to fix issue operator test.  This is definitely what customer was also hitting when attempting to work around the issue.

vr4manta:
PR that addresses the e2e-vsphere-operator issue: 
https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/288

vr4manta:
@damdo Excellent.  Looks like both fixes are working.

damdo:
Indeed, thanks fo the work on this.
/unhold

PR #288, opened 2024-04-16
OCPBUGS-32295: Modified webhook to allow templates by name instead of just by path.

vr4manta:
[OCPBUGS-32295](https://issues.redhat.com/browse/OCPBUGS-32295)

### Changes
- Modified webhook logic when validating CPMS changes to allow vSphere template to be non path formatted as well.

### DIFF
--- a/pkg/webhooks/controlplanemachineset/webhooks.go
+++ b/pkg/webhooks/controlplanemachineset/webhooks.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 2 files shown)


pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -439,12 +440,14 @@ func validateOpenShiftOpenStackProviderConfig(parentPath *field.Path, providerCo
...
-// This ensure that the ControlPlaneMachineSet can safely replace VSphere control plane machines.
+// This ensures that the ControlPlaneMachineSet can safely replace VSphere control plane machines.
 func validateOpenShiftVSphereProviderConfig(parentPath *field.Path, providerConfig providerconfig.VSphereProviderConfig) []error {
 	errs := []error{}
 
+	// Check template format.  If "/" found, assume it's a path based definition and verify with pattern.
+	// If not a path template, then assume just a template name (older ocp installs)
 	templatePath := providerConfig.Config().Template
-	if len(templatePath) > 0 {
+	if len(templatePath) > 0 && strings.Contains(templatePath, "/") {
```
damdo, line 450: Could there ever be a template name that contains `/`?
Are paths _always_ starting with `/`? If so, could use StartsWith instead?

vr4manta: Good question.  I asked Richard and it should start w/ a "/".  If it does not, I can enhance this to return an error for not starting w/ a "/" but containing one.  What are your thoughts on this?

vr4manta: So with the way its coded now, it checks to see if a `/` exists.  If so, it will use the template pattern to make sure it starts with the `/`.  So the current code will check for both.  I think I like keeping it "contains" to allow for catching if format is wrong (where `/` is not at beginning, etc).  Thoughts?

damdo: Ok I see! Then keeping it this way SGTM 👍 thanks for checking this.

damdo (commented):
/approve

PR #289, opened 2024-04-17
[Release 4.15] OCPBUGS-32357: Modified webhook to allow templates by name instead of just by path.

vr4manta:
Cherry-Pick of https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/288

### Changes
- Modified webhook logic when validating CPMS changes to allow vSphere template to be non path formatted as well.


damdo (commented):
/approve

PR #295, opened 2024-05-21
OCPCLOUD-2600: Bump deps to k8s v1.30, go 1.22

damdo:
Bump deps to k8s v1.30, go 1.22


Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.21-openshift-4.16 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.16 AS builder
```
JoelSpeed, line 1:
```suggestion
FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.17 AS builder
```

JoelSpeed (commented):
Looks good but I think you need to update the image to 4.17

Dockerfile
damdo: Yeah, missed it, TY.

.golangci.yaml
```diff
@@ -38,7 +38,6 @@ linters:
     - goimports
     - goprintffuncname
     - gosec
-    - ifshort
```
JoelSpeed, line 41: Was this deprecated?

damdo: Yes

JoelSpeed:
Assuming tests start going green, LGTM

PR #305, opened 2024-06-13
OCPBUGS-35416: Wait for ControlPlaneMachineSet to be created when waiting for it to be updated

JoelSpeed:
Seeing this in presubmits sometimes. One of the tests using this helper expects the CPMS generator to regenerate the CPMS right at the end of the test, but we then immediately call `EnsureControlPlaneMachineSetUpdated` which expects the CPMS to exist.

In the right conditions, this second helper can catch the generator mid way between delete and create as it regenerates, and that causes the test to fail with a not found error.

This should help to prevent that.

### DIFF
--- a/test/e2e/helpers/controlplanemachineset.go
+++ b/test/e2e/helpers/controlplanemachineset.go
@@ -230,11 +230,10 @@ func EnsureControlPlaneMachineSetUpdated(testFramework framework.Framework) {
 
 	Expect(testFramework).ToNot(BeNil(), "test framework should not be nil")
 
-	k8sClient := testFramework.GetClient()
 	ctx := testFramework.GetContext()
 
-	cpms := &machinev1.ControlPlaneMachineSet{}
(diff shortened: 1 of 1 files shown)


test/e2e/helpers/controlplanemachineset.go
```diff
@@ -233,7 +233,9 @@ func EnsureControlPlaneMachineSetUpdated(testFramework framework.Framework) {
 	k8sClient := testFramework.GetClient()
 	ctx := testFramework.GetContext()
 
-	cpms := &machinev1.ControlPlaneMachineSet{}
+	cpms := testFramework.NewEmptyControlPlaneMachineSet()
+	Eventually(komega.Get(cpms)).Should(Succeed(), "control plane machine set should exist")
+
 	Expect(k8sClient.Get(testFramework.GetContext(), testFramework.ControlPlaneMachineSetKey(), cpms)).To(Succeed(), "control plane machine set should exist")
```
damdo, line 239: Can/Should we drop this if we have the eventually above?

JoelSpeed: Oops... I had thought I had done so 😅

damdo (commented):
/approve
/lgtm

PR #120, opened 2022-10-17
OCPBUGS-2469: Don't set a name in related objects

JoelSpeed:
When running `oc adm must-gather` we do not download controlplanemachineset resources.

Must gather uses `oc adm inspect` under the hood which looks at cluster operator related resources to identify what it should download. In must-gather, it does this across all namespaces.

When gathering across all namespaces, you cannot set a name for the related object else an error is produced:
```
error: errors occurred while gathering data:
    skipping gathering controlplanemachinesets.machine.openshift.io/cluster due to error: a resource cannot be retrieved by name across all namespaces
```
We currently set the name of the CPMS in our related objects so we do not get it included in the gather.

PR #310, opened 2024-06-18
NO-JIRA: CPMS Cluster Operator should set related objects

JoelSpeed:
There is an expectation that all cluster operators set at least one namespace and one other object as their related objects. This PR updates our CO generation to set the related objects when they are not as expected.

### DIFF
--- a/pkg/controllers/controlplanemachineset/cluster_operator.go
+++ b/pkg/controllers/controlplanemachineset/cluster_operator.go
@@ -24,10 +24,34 @@ import (
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
 	"github.com/openshift/library-go/pkg/config/clusteroperator/v1helpers"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const machineAPINamespace = "openshift-machine-api"
+
+// relatedObjects returns the expected set of related objects for the CPMS Cluster Operator.
+func relatedObjects() []configv1.ObjectReference {
+	return []configv1.ObjectReference{
+		{
+			Group:    "",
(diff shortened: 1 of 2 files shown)


pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -174,3 +205,28 @@ func getStatusForConditionType(conds []configv1.ClusterOperatorStatusCondition,
 
 	return string(configv1.ConditionUnknown)
 }
+
+// relatedObjectsChanged returns true if the related objects on the ClusterOperator have are different from the expected.
```
damdo, line 209:
```suggestion
// relatedObjectsChanged returns true if the related objects on the ClusterOperator are different from the expected.
```

damdo (commented):
/approve
/lgtm

Only a nit

pkg/controllers/controlplanemachineset/cluster_operator.go
```diff
@@ -24,11 +24,36 @@ import (
...
+	return []configv1.ObjectReference{
+		{
+			Group:    "",
+			Resource: "namespaces",
+			Name:     machineAPINamespace,
+		},
+		{
+			Group:     machinev1.GroupName,
+			Resource:  "controlplanemachinesets",
+			Name:      "cluster",
```
wking, line 46: Comparing with [the ones you already set in your ClusterOperator manifest][1], #120 suggests not setting a name here.  Although the `a resource cannot be retrieved by name across all namespaces` being discussed there seems like maybe the problem was not setting `namespace`?  However you go, you probably want the initial-CVO-pre-creation manifest content and this actively-managed-by-your-operator content to pick the same direction.  

[1]: https://github.com/openshift/cluster-control-plane-machine-set-operator/blame/69ab2b776804db8d1baceb26c18cf938498c3f2c/manifests/0000_30_control-plane-machine-set-operator_04_clusteroperator.yaml#L14-L23

JoelSpeed:
Updated now to fix the nit from Dam and to match what we already have in the CVO pre-created manifest

damdo:
LGTM
Will leave to @wking the final label

wking (approved):
/lgtm

PR #316, opened 2024-07-12
OCPBUGS-36462: Add 1 minute grace period before declaring Nodes unmanaged

JoelSpeed:
We are seeing in CI, that occasionally, there is a race between the CPMS finding a Node, and the NodeLink controller linking that node to the Machines.

This change adds a 1 minute grace period, after the Node is created, to allow it to be attached to a Machine before it is considered to be unmanaged.

This should reduce the false positives in going degraded.

### DIFF
--- a/manifests/0000_30_control-plane-machine-set-operator_03_deployment.yaml
+++ b/manifests/0000_30_control-plane-machine-set-operator_03_deployment.yaml
@@ -30,7 +30,7 @@ spec:
         command:
         - "/manager"
         args:
-        - -v=2
+        - -v=4
         - --leader-elect=true
         - --leader-elect-lease-duration=137s
         - --leader-elect-renew-deadline=107s
--- a/pkg/controllers/controlplanemachineset/controller.go
+++ b/pkg/controllers/controlplanemachineset/controller.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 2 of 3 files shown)


JoelSpeed:
The unmanaged nodes part of the bug that I found seems to be solved, but still seeing issues here. Will need to investigate further.

/hold

damdo:
What issues are you seeing still? @JoelSpeed

JoelSpeed:
Reviewing https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-control-plane-machine-set-operator/316/pull-ci-openshift-cluster-control-plane-machine-set-operator-main-e2e-aws-ovn-etcd-scaling/1811762300021903360, I can still see `unavailableReplicas=1` for a period of time while we are waiting for the new Machine to be ready.

I've added increased the log level to try and understand what the CPMS is thinking of the state of the world during this period.

Thinking about what's happening here, we are effectively trying an `OnDelete` update strategy, though the strategy is set to `RollingUpdate`.  For the replica to be considered unavailable, this means that the Node has gone `Ready!=True`, and the new Node has not yet reached `Ready=True`.

The old Node should not be being drained until the etcd quorum guard is removed, so I would not expect it to go unready (ie it's been deleted on the cloud side?) until the new Node is ready.

Next steps once this comes up are just to confirm what CPMS is seeing and then take a look at the etcd operator actions and the Machine drain controller actions and correlate those to the timestamps

JoelSpeed:
Reviewing the most recent run, I can see that the drain on the Node is happening prior to the new Node being created.

In fact, the etcd operator is removing the etcd guard very early on, much earlier than I would have expected.

I think there is in fact a genuine etcd operator bug here.

JoelSpeed:
/close

We concluded this was an etcd operator issue

PR #327, opened 2024-09-17
OCPBUGS-42660: relax validation on delete and if failureDomains not configured

rvanderp3:
**Description:**
- Do not block deletion of finalizers due to spec validation
- For vSphere, do not enforce empty provider spec workspace if failure domains are not configured in cpms

### DIFF
--- a/pkg/webhooks/controlplanemachineset/test_util.go
+++ b/pkg/webhooks/controlplanemachineset/test_util.go
@@ -28,15 +28,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// createvSphereCPMSWithZones creates a vSphere CPMS and control plane machines which are in corresponding zones.
-func createvSphereCPMSWithZones(
+// createVSphereControlPlaneMachinesWithZones creates control plane machines which are in corresponding failure domains and returns those failure domains.
+func createVSphereControlPlaneMachinesWithZones(
 	ctx context.Context,
 	k8sClient client.Client,
 	namespaceName string,
-	infrastructure *configv1.Infrastructure) *machinev1.ControlPlaneMachineSet {
-	providerSpec := machinev1beta1resourcebuilder.VSphereProviderSpec().AsControlPlaneMachineSetProviderSpec()
-	machineTemplate := machinev1resourcebuilder.OpenShiftMachineV1Beta1Template().WithProviderSpecBuilder(providerSpec)
-
+	infrastructure *configv1.Infrastructure) *machinev1.FailureDomains {
 	fdBuilder := machinev1resourcebuilder.VSphereFailureDomains()
 	failureDomains := fdBuilder.BuildFailureDomains()
 
@@ -47,13 +44,8 @@ func createvSphereCPMSWithZones(
 		controlPlaneMachineBuilder := machineBuilder.WithGenerateName("control-plane-machine-").AsMaster().WithProviderSpecBuilder(machineProviderSpec)
 
 		controlPlaneMachine := controlPlaneMachineBuilder.Build()
-		Expect(k8sClient.Create(ctx, controlPlaneMachine)).To(Succeed(), "expected to be able to create the control plane machine")
+		Eventually(k8sClient.Create(ctx, controlPlaneMachine)).Should(Succeed(), "expected to be able to create the control plane machine")
 	}
 
-	cpmsBuilder := machinev1resourcebuilder.ControlPlaneMachineSet().WithNamespace(namespaceName).WithMachineTemplateBuilder(machineTemplate)
-	cpms := cpmsBuilder.Build()
-
-	cpms.Spec.Template.OpenShiftMachineV1Beta1Machine.FailureDomains = &failureDomains
-
-	return cpms
+	return &failureDomains
 }
--- a/pkg/webhooks/controlplanemachineset/test_util.go
+++ b/pkg/webhooks/controlplanemachineset/test_util.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controlplanemachineset
+
+import (
+	"context"
+
+	. "github.com/onsi/gomega"
+
+	configv1 "github.com/openshift/api/config/v1"
+	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1"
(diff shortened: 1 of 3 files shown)


pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -142,6 +142,20 @@ func (r *ControlPlaneMachineSetWebhook) ValidateCreate(ctx context.Context, obj
 	return warnings, nil
 }
 
+// ShouldValidateSpecUpdate determines if the spec should be validated. For vSphere, in particular, cpms's created before 4.16
+// may have fields that no longer pass validatation due to using the infrastructure resource as a configuration source. as such,
+// it is possible to have the cpms be configured such that it can't be deleted as validation blocks removal of finalizers.
+func (r *ControlPlaneMachineSetWebhook) ShouldValidateSpecUpdate(ctx context.Context, oldCpms, newCpms *machinev1.ControlPlaneMachineSet) (bool, error) {
```
damdo, line 148: I think this might not need to be exported

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -142,6 +142,20 @@ func (r *ControlPlaneMachineSetWebhook) ValidateCreate(ctx context.Context, obj
...
+	// if cpms is being deleted and finalizers are being dropped, do not validate further
```
damdo, line 149:
```suggestion
	// If cpms is being deleted and finalizers are being dropped, do not validate further.
```

pkg/webhooks/controlplanemachineset/webhooks.go
damdo: Probably also this can be a helper function without being a method on the ControlPlaneMachineSetWebhook receiver, as it doesn't make use of it

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -142,6 +142,20 @@ func (r *ControlPlaneMachineSetWebhook) ValidateCreate(ctx context.Context, obj
...
+	if newCpms.DeletionTimestamp != nil {
+		if len(oldCpms.Finalizers) > len(newCpms.Finalizers) && len(newCpms.Finalizers) == 0 {
+			return false, nil
+		}
+	}
+
+	return true, nil
```
damdo, line 156: error is always nil here, should we drop it?

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -152,11 +166,25 @@ func (r *ControlPlaneMachineSetWebhook) ValidateUpdate(ctx context.Context, oldO
...
 	}
 
+	shouldValidate, err := r.ShouldValidateSpecUpdate(ctx, oldCpms, cpms)
+	if err != nil {
+		return warnings, fmt.Errorf("error checking if spec should be validated: %w", err)
+	}
+
+	if !shouldValidate {
+		return warnings, nil
+	}
```
damdo, line 186: If we drop the error from the return we could do something like:
```
if !shouldValidateSpecUpdate(oldCpms, cpms) {
  return warnings, nil
}
```

pkg/webhooks/controlplanemachineset/webhooks.go
```diff
@@ -491,6 +521,15 @@ func checkForFailureDomainIgnoredFields(parentPath *field.Path, providerConfig p
 		errs = append(errs, validateTemplate(parentPath, providerConfig.Config().Template)...)
 	}
 
+	// if a cluster born before 4.16 has a configured cpms that cpms may(and likely will) contain fields that
+	// conflict with the model of using the infrastructure resource as the configuration source.
+	// in those situations, it is necessary to only validate for field conflicts if failure
+	// domains are defined in the cpms. otherwise, the cpms could be unmodifiable/undeletable due to it
+	// failing validation.
```
damdo, line 528:
```suggestion
	// If a cluster born before OCP 4.16 has a configured cpms that cpms may(and likely will) contain fields that
	// conflict with the model of using the infrastructure resource as the configuration source.
	// In those situations, it is necessary to only validate for field conflicts if failure
	// domains are defined in the cpms, otherwise, the cpms could be unmodifiable/undeletable due to it
	// failing validation.
```

pkg/webhooks/controlplanemachineset/test_util.go
```diff
@@ -0,0 +1,59 @@
...
+	failureDomains := fdBuilder.BuildFailureDomains()
+
+	for _, fd := range failureDomains.VSphere {
+		machineProviderSpec := machinev1beta1resourcebuilder.VSphereProviderSpec().WithInfrastructure(*infrastructure).WithZone(fd.Name)
+
+		machineBuilder := machinev1beta1resourcebuilder.Machine().WithNamespace(namespaceName)
+		controlPlaneMachineBuilder := machineBuilder.WithGenerateName("control-plane-machine-").AsMaster().WithProviderSpecBuilder(machineProviderSpec)
+
+		controlPlaneMachine := controlPlaneMachineBuilder.Build()
+		Expect(k8sClient.Create(ctx, controlPlaneMachine)).To(Succeed(), "expected to be able to create the control plane machine")
```
damdo, line 50: Better to use Eventually here to tolerate temporary API server blips

pkg/webhooks/controlplanemachineset/test_util.go
```diff
@@ -0,0 +1,59 @@
...
+
+	configv1 "github.com/openshift/api/config/v1"
+	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1"
+	machinev1beta1resourcebuilder "github.com/openshift/cluster-api-actuator-pkg/testutils/resourcebuilder/machine/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// createvSphereCPMSWithZones creates a vSphere CPMS and control plane machines which are in corresponding zones.
+func createvSphereCPMSWithZones(
```
damdo, line 32: I think we could improve the name of this to capture more precisely what it does.
In here we are **creating** control plane machines, and we are **generating** the CPMS object, which will instead get actually **created** in the cluster later.

Maybe we could call it `createVSphereControlPlaneMachinesWithZones()`, only create CP machines, return  the FailureDomains and instead generate the CPMS outside of this one, in a different helper which we can call subsequently to this in tests.

damdo (commented):
Mostly LGTM, left some nits :)

damdo (commented):
/approve
/lgtm

damdo:
I think the etcd scaling failure is unrelated to this change as that runs on AWS only.

We might want to chat with the etcd folks about that one.

rvanderp3:
> I think the etcd scaling failure is unrelated to this change as that runs on AWS only.
> 
> We might want to chat with the etcd folks about that one.

yeah, i'll start a thread with them

damdo:
@rvanderp3  I already did here, it seems like it is a known issue they are trying to fix: https://redhat-internal.slack.com/archives/C027U68LP/p1730229820314459

damdo:
@rvanderp3 let me know if you want to wait for https://github.com/openshift/cluster-kube-storage-version-migrator-operator/pull/118 to merge or if we should override that scaling job

rvanderp3:
> @rvanderp3 let me know if you want to wait for [openshift/cluster-kube-storage-version-migrator-operator#118](https://github.com/openshift/cluster-kube-storage-version-migrator-operator/pull/118) to merge or if we should override that scaling job

I'm fine with overriding.  It would be good to get this in sooner rather than later i think.

PR #328, opened 2024-09-24
OCPCLOUD-2734: Bump deps to k8s v1.31

damdo:
Bump deps to k8s v1.31


Makefile
```diff
@@ -1,5 +1,5 @@
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.29.1
+ENVTEST_K8S_VERSION = 1.31.0
```
JoelSpeed, line 2: We don't currently publish openshift versions for this release. Since this doesn't exist in the bucket, it will fall back to the defaults, which should be ok.

Can you update the envtest invocation to remove the remote bucket since we no longer need to rely on OCP forks

damdo: Ah yes, good point, we are still running off of the OCP fork bucket.

Now removed, so we will hit the upstream bucket, which has 1.31.0.

nrb:
Nit: commit message references k8s v1.30 when the actual change is v1.31

damdo:
@nrb fixed thanks. Could you re-add the LGTM label please? Thanks!

damdo:
/test unit
Flaked

PR #350, opened 2025-02-26
Workaround static pod controller bug by doubling timeout

mdbooth:
The periodics are consistently hitting
https://issues.redhat.com/browse/OCPBUGS-50587, manifesting as a timeout
waiting for the cluster to stabilise in the MachineNamePrefix test. This
change simply doubles the timeout, as the sitution should resolve itself
eventually. We should remove this workaround when the bug is fixed.

### DIFF
--- a/test/e2e/helpers/cases.go
+++ b/test/e2e/helpers/cases.go
@@ -173,7 +173,10 @@ func ItShouldRollingUpdateReplaceTheOutdatedMachine(testFramework framework.Fram
 		By("Control plane machine rollout completed successfully")
 
 		By("Waiting for the cluster to stabilise after the rollout")
-		EventuallyClusterOperatorsShouldStabilise(1*time.Minute, 30*time.Minute, 30*time.Second)
+		// Double the timeout to 60 minutes as a temporary workaround for
+		// https://issues.redhat.com/browse/OCPBUGS-50587. Put this back to a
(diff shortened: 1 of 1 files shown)


mdbooth:
We expect to see openstack-periodic-pre-techpreview fail until https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/349 merges, but I've run it for completeness.

JoelSpeed:
I don't think the fix ever resolves itself does it? My understanding was that it is seeing nodes that have been removed, even after they're removed, and it never times out?

mdbooth:
> I don't think the fix ever resolves itself does it? My understanding was that it is seeing nodes that have been removed, even after they're removed, and it never times out?

My reading of the bug is that it's a race that eventually sorts itself when the updates occur in the happy order.

mdbooth:
OpenStack failed during installation, so unrelated.

The azure test actually passed, but hit the 4h global timeout during cleanup.

This looks like a reasonably successful bodge to me. However, as https://github.com/openshift/api/pull/2208 merged overnight which hopefully fixes this properly it's probably not worth it. If we haven't decided to merge this PR within the next couple of days we should close it.

/hold

mdbooth:
https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/349 seemed to pass pretty quickly. I'm going to optimistically assume that this is fixed.

PR #338, opened 2025-01-07
OAPE-78: Add e2e periodic job for machine name prefix

chiragkyal:
- This is required to gather test data for https://github.com/openshift/enhancements/pull/1714 feature.
- Techpreview periodics job: https://github.com/openshift/release/pull/60342 , https://github.com/openshift/release/pull/60377

- Part of: https://issues.redhat.com/browse/OAPE-78

PR #332, opened 2024-11-06
OAPE-19: Support customized control plane machine names with prefix

chiragkyal:
This PR introduces the ability to customize the naming format of Control Plane Machines. This is achieved through a new `machineNamePrefix` field in the ControlPlaneMachineSet spec, gated behind the `CPMSMachineNamePrefix` feature gate.

When feature gate is enabled and `machineNamePrefix` field is specified, it allows users to specify a prefix that will be used in `getMachineName`. The resulting machine name will consist of the provided prefix, followed by a randomly generated string of 5 characters and the machine index.

Otherwise, it will fall back to the default naming convention.

- Implementes : https://github.com/openshift/enhancements/pull/1714
- API PR: https://github.com/openshift/api/pull/2086
- Part of: [OAPE-19](https://issues.redhat.com//browse/OAPE-19)

PR #353, opened 2025-03-04
OAPE-147: Add E2E tests for MachineNamePrefix with OnDelete update strategy

chiragkyal:
This PR adds E2E tests (both pre-submit and periodic) for `MachineNamePrefix` feature with `OnDelete` update strategy. 

The existing tests work only for `RollingUpdate` strategy. This PR extends the tests added through:

- https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/338
- https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/332

Part of : https://issues.redhat.com/browse/OAPE-147

### DIFF
--- a/test/e2e/periodic_test.go
+++ b/test/e2e/periodic_test.go
@@ -24,6 +24,8 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 
 	"github.com/openshift/api/features"
+	machinev1 "github.com/openshift/api/machine/v1"
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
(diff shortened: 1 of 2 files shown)


chiragkyal:
/assign @JoelSpeed 

After this PR, we will have two e2e tests for this feature in total, which is lesser than the required 5 tests for promoting a feature to GA. I think we need to make an exception here if we cannot think of any other potential scenarios.

chiragkyal:
The `e2e-aws-ovn-etcd-scaling` is really having a hard time. 

/retest

JoelSpeed:
The test failed because there are unexpected transitions for the controlplanemachineset operator, would you mind spending some time to see if you can explain why the CPMS is transitioning to unavailable?

PR #356, opened 2025-04-03
OCPBUGS-52448: Remove gathering of failure domains from machine sets

RadekManak:
This fixes a bug when a cluster is running with 3 control plane nodes in a single AZ, and machine pools in > 1 AZ, CPMS does not generate a config. 

We decided to remove the feature that gathers additional failure domains from MachineSets. While useful, this feature prevents the generation of the CPMS in the case mentioned above. Our priority is to generate a valid CPMS based on the current state of the control plane, allowing the cluster administrator to add failure domains later if needed, rather than requiring manual intervention upfront.

### DIFF
--- a/pkg/controllers/controlplanemachinesetgenerator/controller_test.go
+++ b/pkg/controllers/controlplanemachinesetgenerator/controller_test.go
@@ -265,7 +265,7 @@ var _ = Describe("controlplanemachinesetgenerator controller on AWS", func() {
 						WithMachineTemplateBuilder(
 				machinev1resourcebuilder.OpenShiftMachineV1Beta1Template().
 					WithProviderSpecBuilder(
-						usEast1aProviderSpecBuilderAWS.WithAvailabilityZone("").WithSubnet(machinev1beta1.AWSResourceReference{}).WithInstanceType("c5.8xlarge"),
+						usEast1aProviderSpecBuilderAWS.WithAvailabilityZone("").WithSubnet(machinev1beta1.AWSResourceReference{}).WithInstanceType("c5.4xlarge"),
 					).
 					WithFailureDomainsBuilder(machinev1resourcebuilder.AWSFailureDomains().WithFailureDomainBuilders(
 						usEast1aFailureDomainBuilderAWS,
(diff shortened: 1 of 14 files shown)


JoelSpeed (commented):
Changes make sense, but we will need to get the tests updated

pkg/controllers/controlplanemachineset/controller_test.go
```diff
@@ -1640,7 +1640,7 @@ var _ = Describe("With a running controller and machine name prefix", func() {
 			})
 
 			It("should keep the status unchanged consistently", func() {
-				Consistently(komega.Object(cpms)).Should(HaveField("Status", SatisfyAll(
+				Consistently(komega.Object(cpms), 1*time.Second).Should(HaveField("Status", SatisfyAll(
```
JoelSpeed, line 1643: What is our default consistently timeout? Does the suite set this somewhere?

pkg/controllers/controlplanemachinesetgenerator/controller_test.go
```diff
@@ -378,9 +353,15 @@ var _ = Describe("controlplanemachinesetgenerator controller on AWS", func() {
 		// Create Machines with some wait time between them
 		// to achieve staggered CreationTimestamp(s).
 		Expect(k8sClient.Create(ctx, machine0)).To(Succeed())
+		time.Sleep(1 * time.Second)
```
JoelSpeed, line 356: Did this not work before? I see the comment above suggests there was already wait time

RadekManak: The machines had the same creationTimestamp. The test still pased becuase of the short duration of Consistently interval and because we sort machines with the same timestamp by name.

RadekManak (commented):
Set default consistently timeout and fixed platformNone tests that the change revealed to be broken.

pkg/controllers/controlplanemachineset/controller_test.go
RadekManak: There was no default set. The gomega default was 100ms with polling interval of 10ms.

I have retested this change and changed the default to 500ms with 50ms pooling interval in the unit controlplanemachineset, controlplanemachinesetgenerator and machine provider test suite.

After this change, I found the Noneplatform test was also broken.

RadekManak:
/hold
I think this might be braking e2e-aws-operator-techpreview. I'll investigate.

damdo:

This is a known issue and not related to this PR (we have a potential fix for it on #357)

PR #357, opened 2025-04-07
WIP: add unmanaged grace period for new node

damdo:
Add unmanaged grace period for new node to avoid going immediately degraded if a node has just been created and hasn't been linked to a machine yet.


JoelSpeed:
Conceptually this makes sense to me to avoid rapid flicks of degraded/stable

/approve

PR #358, opened 2025-04-23
OCPBUGS-55017: Wait for operators to be stable for 2 minutes before accepting cluster is stable

JoelSpeed:
We are seeing that the 1 minute period is not enough to pick up etcd operator rolling out several revisions.

Increase the stabilisation period to give a 2 minute stable requirement before we accept that the cluster has stabilised


JoelSpeed:

This has been permafailing for ages and won't be affected by the changes in this PR

PR #364, opened 2025-07-29
Disable activation of CPMS with deletionTimestamp

RadekManak:
This should prevent e2e-aws-operator-techpreview from failing by rejecting activation of CPMS that is to be deleted.

There is an underlying issue with tech preview that is surfacing this where cpms is outdated due to
Template.OpenShiftMachineV1Beta1Machine.Spec.AuthoritativeAPI: MachineAPI != "".

I'll fix that in another PR to independently verify both fixes.


damdo:
Nice catch @RadekManak 
Can you please create a bug for this? TY

damdo (commented):
/approve
/lgtm

/hold

for testing. Feel free to remove when happy.

RadekManak:
/unhold
I have seen the webhook successfully block the activation before #365 merged.

PR #365, opened 2025-07-29
OCPBUGS-59919: Fix authoritativeAPI recreate loop

RadekManak:
This PR resolves a bug where generatedCPMS did not set AuthoritativeAPI. This omission caused continuous diffs against the default value, trapping CPMS in a persistent delete-create loop.

The fix implements two changes:
1.  AuthoritativeAPI is now populated from the latest control plane machine's value.
2. Diff checking for AuthoritativeAPI has been disabled as it should not cause machine replacement.


damdo (commented):
/approve

Thanks!

RadekManak:
/override ci/prow/e2e-aws-ovn-etcd-scalingf
This test is known to not pass.

